PR #93, opened 2019-02-12
parser updates and general optimizations

brannon-darby:


brannon-darby:
fixes https://github.com/prisma/nexus-prisma/issues/91

Weakky:
Hey, thanks for this PR ! Everything looks good!

As `prisma-datamodel` is still in alpha, breaking changes are expected without notice.

Maybe we should fix the version of `prisma-datamodel` so that it doesn't happen again in the future.

Merging now 🙌

PR #94, opened 2019-02-12
fix(npg): generated metaSchema export

brannon-darby:


brannon-darby:
named exports are the way to go in ts - default exports tend to cause problems (like this one and the one related to my previous pr).

brannon-darby:
my mistake - the issue was with my own updated schema locally.  i can see why you might want to make that the default nexus export now.  closing.

PR #97, opened 2019-02-14
[RFR] BREAKING CHANGE: `t.prismaFields()`

Weakky:
- [x] 1) A `definition` function is always required
- [x] 2) An empty `definition` function exposes no fields
- [x] 3) We introduce `*` as a field option to explicitly expose all fields via `t.prismaFields`
- [x] 4) We require an array argument in the `t.prismaFields([...])` function. An empty array means no prisma fields exposed
- [x] 5) Rename `metaSchema` to `datamodelInfo`


README.md
```diff
@@ -108,64 +97,137 @@ interface Field {
...
+  name: 'User',
+  definition(t) {
+    t.prismaFields({ filter: ['id', 'name'] })
+  },
+})
+```
 
-export const Mutation = prismaObjectType({
-  name: 'Mutation',
+**Exposes the only the `users` field, and alias it to `customers`**
```
t0wer001:
```suggestion
**Exposes only the `users` field, and alias it to `customers`**
```

Weakky: Thanks for the fix!

jeremyadavis:
Just an FYI, in `makePrismaSchema` the change from `metaSchema` to `datamodelInfo` breaks the example used in the [Using GraphQL Nexus with a Database](https://www.prisma.io/blog/using-graphql-nexus-with-a-database-pmyl3660ncst/) tutorial.

jaredjasonpeters:
### Rename the default export to datamodelInfo and structure it this way

```
prisma: {
datamodelInfo,
client: prisma
} 
```

### or just use it as a property and pass it metaSchema

```const schema = makePrismaSchema({
  types: [Query, Mutation],

  prisma: {
    datamodelInfo: metaSchema,
    client: prisma
  },

  outputs: {
    schema: path.join(__dirname, "./generated/schema.graphql"),
    typegen: path.join(__dirname, "./generated/nexus.ts")
  }
});

jeremyadavis:
Yeah, I fixed it and got through the rest of the tutorial without issue. Just thought you'd like to know to update the tutorial. Sorry, didn't clarify that.

PR #114, opened 2019-02-17
Document prismaExtendType

jangerhofer:
I spent a bit of time grappling with the methodology for splitting up the fields of my `Query` and `Mutation` types across different modules/files, while still maintaining access to my Prisma models.  After I read this [great writeup in the Nexus docs](https://nexus.js.org/docs/database-access-with-prisma), I thought to look for the `nexus-prisma` equivalent of `extendType`.  Lo and behold, VSCode immediately suggested `prismaExtendType` (thank you, IntelliSense!).

I want to save others that effort of searching for the right mechanism, so I wrote up an outline of the function and provided context for its usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,9 @@ hooks:
 
 ### `prismaObjectType()`
 
-`prismaObjectType` is a wrapper around Nexus' `objectType`. It expects an object with the following properties:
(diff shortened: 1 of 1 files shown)


Weakky:
Hey there,

Our docs are still missing some important part so many thanks for taking the time to document that function!

Merging now 🙌

PR #272, opened 2019-05-11
Use given path to prisma.yml when possible

denkristoffer:
I wasn't able to get `nexus-prisma-generate` 0.3.7 to run because the `findPrismaConfigFile` function as it is ignores the given `--project` flag and overwrites the path. This PR is a fix.


Weakky:
Hey there 👋,

Thanks a lot for this fix, merging now 🙏

danny-p93:
could you update the public npm package because that still doesn't contains the fix.

janpio:
That would mean a new release 0.3.8 @danny-p93 - you could probably create a new issue to request this.

danny-p93:
@janpio thanks for the hint, i created one

PR #385, opened 2019-09-03
fix: enums

### LINKED ISSUE #323: enum types only generated for mutations
enum types are not generate if they are not in an mutation type. In my example the model Profile can have a list of DeactivatedPush. After I created the first mutation updateOneProfile the enum type is imported correctly.

Before adding the mutation an error was thrown:
- Missing type PushSetting, did you forget to import a type to the root query?

After adding the mutation:
[ERROR] 14:00:17 Error: PushSetting was already defined and imported as a type, check the docs for extending types

in project.prisma I have:

```
enum PushSetting {
  Message
  Gift
  Friendship
  Visit
  Poke
  Picture
  SystemMessage
}

model DeactivatedPush {
  id      Int         @id
  setting PushSetting @default(SystemMessage)
}

```

nexus do not generate this type so I have to write it again

```
import { objectType, enumType } from 'nexus';
import { queries, QueryMethod } from './';

queries.push({ name: "DeactivatedPush", methods: [QueryMethod.Many, QueryMethod.One] });

export const DeactivatedPush = objectType({
    name: 'DeactivatedPush',
    definition(t) {
        t.model.id()
        t.model.setting()
    },
});

export const PushSetting = enumType({
    name: "PushSetting",
    members: {
        Message: "Message",
        Gift: "Gift",
        Friendship: "Friendship",
        Visit: "Visit",
        Poke: "Poke",
        Picture: "Picture",
        SystemMessage: "SystemMessage"
    }
});
```

In the genarated Photon file I have the type

```
export declare const PushSetting: {
    Message: "Message";
    Gift: "Gift";
    Friendship: "Friendship";
    Visit: "Visit";
    Poke: "Poke";
    Picture: "Picture";
    SystemMessage: "SystemMessage";
};
export declare type PushSetting = (typeof PushSetting)[keyof typeof PushSetting];

```
Slack link: https://prisma.slack.com/archives/CKQTGR6T0/p1562188139286800

### LINKED ISSUE #383: Enums cannot be exposed via model
Enums work well now both for mutations and queries.

However, the typegen still seems to be missing something.

In the definition of the field (`t.model.myfield`), I get the following error:
```
ERROR: No subset types are available. Please make sure that one of your GraphQL type is a subset of your t.model('<ModelName>')
```
Screenshot:
Screenshot 2019-09-03 at 07 17 45

jasonkuhrt:
Will fix #383  
Will fix #323

### DIFF
--- a/tests/__snapshots__/schema.test.ts.snap
+++ b/tests/__snapshots__/schema.test.ts.snap
@@ -1,5 +1,20 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`enum can be filtered on 1`] = `
+"type A {
+  a: Int!
+}
+
+input AWhereUniqueInput {
+  a: Int
+}
+
+type Query {
+  a(where: AWhereUniqueInput!): A
+}
+"
+`;
+
 exports[`it exposes filtering only if filtering: true 1`] = `
 "input IntFilter {
   equals: Int
--- a/tests/__snapshots__/schema.test.ts.snap
+++ b/tests/__snapshots__/schema.test.ts.snap
@@ -1,16 +1,24 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`enum can be filtered on 1`] = `
-"type A {
-  a: Int!
+"enum EA {
+  A
+  B
+  C
 }
 
-input AWhereUniqueInput {
-  a: Int
+type MA {
+  fa: Int!
+  fb: EA!
+}
+
+input MAWhereUniqueInput {
+  fa: Int
 }
 
 type Query {
-  a(where: AWhereUniqueInput!): A
+  ma(where: MAWhereUniqueInput!): MA
+  mas(skip: Int, after: String, before: String, first: Int, last: Int): [MA!]
 }
 "
 `;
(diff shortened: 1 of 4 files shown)


tests/__snapshots__/schema.test.ts.snap
```diff
@@ -1,5 +1,28 @@
...
+  C
+}
+
+type MA {
+  fa: Int!
+  fb: EA!
+}
+
+input MAWhereUniqueInput {
+  fa: Int
```
jasonkuhrt, line 16: @Weakky 

1. The test and [the example](https://github.com/prisma/nexus-prisma/pull/385/files#diff-97a48ec5ce235535fdb2c30a627f12e7R210) are producing different results 🤔 
2. I think `Int` is wrong here. For example `fa: EA` would make a lot more sense? I have a feeling this `Int` comes from the backing prisma enum (which is an enum in sense of names over ints).

tests/schema.test.ts
```diff
@@ -207,3 +207,35 @@ test('it exposes findOne and findMany', async () => {
...
+    },
+  })
+  const MA = objectType({
+    name: 'MA',
+    definition(t: any) {
+      t.model.fa()
+      t.model.fb()
+    },
+  })
+  const EA = enumType({ name: 'EA', members: ['A', 'B', 'C'] })
```
jasonkuhrt, line 236: @Weakky you mentioned this shouldn't be needed any more right? But if we remove this, the test will fail.

PR #386, opened 2019-09-04
refactor: build crud

jasonkuhrt:


src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -22,8 +17,9 @@ import {
 } from './NamingStrategies'
 import { dateTimeScalar, GQL_SCALARS_NAMES } from './scalars'
 import { getSupportedMutations, getSupportedQueries } from './supported-ops'
+import { DynamicOutputPropertyDef } from 'nexus/dist/dynamicProperty'
 
-interface NexusPrismaMethodParams {
+interface FieldPublisherConfig {
```
jasonkuhrt, line 22: I think we need a good name for what this part of the system is. I considered `Field Configurator` but then preferred `Field Publisher` because `publisher` has connotations of "exposing" something (e.g. docker port publishing) and `publish` is a much less generic/overloaded term than `config`.

`Field` is still generic here. I think we could do better. But not sure, this seems like a good step already.

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -64,32 +60,23 @@ export class NexusPrismaBuilder {
...
-
     return dynamicOutputProperty({
       name: 'crud',
       typeDefinition: `: NexusPrisma<TypeName, 'crud'>`,
       factory: ({ typeDef: t, typeName: graphQLTypeName }) => {
-        if (graphQLTypeName !== 'Query' && graphQLTypeName !== 'Mutation') {
-          throw new Error(
-            `t.crud can only be used on a 'Query' & 'Mutation' objectType. Please use 't.model' instead`,
-          )
-        }
```
jasonkuhrt, line 89: guard is now a bit lower in an else clause. I like guards, but the pattern seems a bit spontaneous here and felt it became simpler to use `else` after the short-circuit returns were removed.

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -64,32 +60,23 @@ export class NexusPrismaBuilder {
...
+        let mappedFields: FieldsWithModelName[]
```
jasonkuhrt: Aside: Have not fully wrapped my head around `mappedFields`. I started reading the dmmf module tonight but still hazy.

jasonkuhrt: Instead of early return we do branch assignment and then continue working within this function.

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -112,17 +95,80 @@ export class NexusPrismaBuilder {
...
           })
-
-          return this.buildSchemaForCRUD(t, 'Mutation', mutationFields)
+        } else {
+          throw new Error(
+            `t.crud can only be used on a 'Query' & 'Mutation' objectType. Please use 't.model' instead`,
+          )
         }
+
+        type FieldPublisher = (opts?: FieldPublisherConfig) => CRUDMethods // Fluent API
```
jasonkuhrt, line 105: captured the fluent api in the type system

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -281,71 +327,6 @@ export class NexusPrismaBuilder {
     }, {})
   }
 
-  protected buildSchemaForCRUD(
```
jasonkuhrt, line 284: Bringing this inline removes a level of indirection. I am struggling to follow the layers and didn't feel this one brought much value.

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -112,17 +95,80 @@ export class NexusPrismaBuilder {
...
+                key => (mappedField.mapping as any)[key] === field.name,
+              ) as keyof DMMF.Mapping | undefined
+
+              if (!operationName) {
+                throw new Error(
+                  `Could not find operation name for field ${field.name}`,
+                )
+              }
+
+              t.field(gqlFieldName, {
```
jasonkuhrt, line 138: I removed use of the helper func and spread here. Felt it was easier to read direct usage of the nexus api rather than hiding just a few fields behind spread. We can afford to lose some DRY so long as the type system keeps us honest.

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -112,17 +95,80 @@ export class NexusPrismaBuilder {
...
+          )
         }
+
+        type FieldPublisher = (opts?: FieldPublisherConfig) => CRUDMethods // Fluent API
+        type CRUDMethods = Record<string, FieldPublisher>
+
+        return mappedFields.reduce<CRUDMethods>((crud, mappedField) => {
+          const prismaModelName = mappedField.mapping.model
+
+          mappedField.fields.forEach(field => {
```
jasonkuhrt, line 111: I expect that more refactoring will come down on this forEach block and all its contents as I get a better handle on the systems/layers at play.

I suspect it would help a lot if we could introduce a few maps instead of one giant forEach. Worse for performance (if eager) but easier to reason about––but not sure, maybe teasing apart a few discrete stages here will prove illusive.

src/nexus-prisma/NexusPrismaBuilder.ts
Weakky: Given that we're in the crud block, and that we know for a fact that it'll only be available in the `Query` and `Mutation` type, I generally like to make that kind of guard high so that it's very clear, and also to make sure that we won't have to care about it anytime later in the code. I guess that's about taste, I'm fine with that too!

src/nexus-prisma/NexusPrismaBuilder.ts
Weakky: In the DMMF format, `mappings` is an array of object which maps for each model in the prisma schema, the type of operation to the corresponding graphql field name. By type of operation, I mean `findOne`, `findMany`, `create`, `update`, etc.
In other words, if you have a `Blog` model in your schema, mapping will look something like this:
```js
{
 "model": "Blog",
 "findOne": "blog",
 "findMany": "blogs",
 "create": "createBlog",
 "update": "updateBlog"
}
```
While it might seem like we can compute the name of these fields, the goal is here is to have one single source of truth (DMMF is computed by the core), especially regarding pluralization where bad stuff can happen

As for `mappedFields`, it's just a small datastructure that combine the mappings of a type with the graphql fields information for each of the field names given by the mapping

While I'm pretty sure the underlying logic is good, this definitely requires some refactor/comment to make it clearer

Weakky: > Instead of early return we do branch assignment and then continue working within this function

Same as above with the guards, that's a code-style that I started to use more and more.

- I generally prefer 3 separated if with a clear condition, rather than a if/elseif/elseif/else, which IMO makes it harder to read because of conditions that rely on the previous ones.
You might argue that this pattern can lead to more verbose conditions, and that's true.
I use it when it's not the case, and generally, never use `else` because the ifs before already contain early return, so we can basically just replace the else by a simple return in the function scope.

- I also find early returns to make the code mentally easier, as you can digest the code block by block (because it stops), without having to take into account whatever will happen after

- In this case, it also prevents you from messing with the `mappedFields` variable by mistake (as we used to have two const instead of one let). Let's say you declare another variable lower in the code which starts by `mapped` too, you then start working with it and trust vscode intellisense, but make a small mistake and choose the upper one `mappedFields`

Finally, while this might seem a lot for that simple example, these are some arguments that mostly apply to more complex example, but that I ended up using even for simple cases out of habits

PS: I don't feel necessarily strong about keeping the way it was (although it seems so 😄). I like the fact that we removed `buildSchemaForCrud`. I'll try to develop these kinds of habits/reasoning as we're starting our collaboration so that you have an idea of my programming style and that we slowly find a common ground to code with some norms/same patterns 🙌

src/nexus-prisma/NexusPrismaBuilder.ts
Weakky: Good! I used to have a lot more of these nexus configs in nexus-prisma v1, which lead to some mistakes from time to time where I'd forget one field (like `isList`). We don't have that much anymore, so it's fine that way!

src/nexus-prisma/NexusPrismaBuilder.ts
Weakky: Agreed! As you mentioned before, we can probably refactor the `forEach`, which will simplify things even more

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -64,32 +60,23 @@ export class NexusPrismaBuilder {
...
   }
 
-  getNexusPrismaMethod() {
-    return [
-      this.getCRUDDynamicOutputMethod(),
-      this.getModelDynamicOutputMethod(),
-      ...this.getPrismaScalars(),
-    ]
+  build() {
+    return [this.buildCRUD(), this.buildModel(), ...this.buildScalers()]
```
Weakky, line 64: typo on `Scalers`

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -523,7 +504,7 @@ export class NexusPrismaBuilder {
     return inputTypeName
   }
 
-  protected getPrismaScalars() {
+  protected buildScalers() {
```
Weakky, line 508: Same as above, typo on `Scalers`

src/nexus-prisma/NexusPrismaBuilder.ts
Weakky: The reason why it's a forEach is that we're calling `t.field()`, which we cannot do in something else.
We could, however, have a map that'd precompute all the nexus configs, then have a `forEach` that'd just call the `t.field` and pass the config straight

src/nexus-prisma/NexusPrismaBuilder.ts
Weakky: I really like `publisher`. I don't feel like `field` is being generic here, we're referring to a "GraphQL field", which AFAIK is the right terminology in their spec. Let's keep it that way!

src/nexus-prisma/NexusPrismaBuilder.ts
jasonkuhrt: Yeah I hear you. I'm kind of imagining this could have been done using pattern matching in something like Rust or Haskell, and each branch dispatches except for the `otherwise` where the runtime error is raised. In fact that's close to what you had before!

src/nexus-prisma/NexusPrismaBuilder.ts
jasonkuhrt: > We could, however, have a map that'd precompute all the nexus configs, then have a forEach that'd just call the t.field and pass the config straight

Yeah! Idea there I guess being we have one/a few pure data transformations then the final side-effect

src/nexus-prisma/NexusPrismaBuilder.ts
jasonkuhrt: Thanks for sharing this! Reflected on this and tried a few experiments;

#### if-guard + ternary + never-default

image

#### "pattern-matching" + throw-default

TS limitation: throw default leads TS to see unreachable code

image

#### if-guard + "pattern-matching"

TS limitation: variable is seen as unassigned unless never-default provided

image

Of these, I think `if-guard + ternary + never-default` is best option here. And I think that brings us [more] into alignment too! 🙌

PR #401, opened 2019-09-09
Publish uuid scalar when needed

### LINKED ISSUE #335: How do you get uuid() working?
I'm building out my prisma2 api. We use UUID's instead of cuid's. However, I'm getting the error `- Missing type UUID, did you forget to import a type to the root query?` when `yarn start`. 

Here is an example model: 
```
model User {
  id                  String                @default(uuid()) @id @unique
}
```

Weakky:
Fixes #335

Doesn't fix the `BatchPayload` object output type that is missing when using `t.crud.updateMany` though. That should be on a separate issue


jasonkuhrt (approved):
LGTM

PR #402, opened 2019-09-09
Publish only pagination params that are true

### LINKED ISSUE #376: Prisma2: Pagination boolean doesn't work
### Steps to reproduce
- While following the documentation for [Prisma 2](https://nexus.js.org/docs/database-access-with-prisma-v2) the pagination boolean value for creating an `objectType` doesn't seem to have any effect.

```ts
import { objectType } from '@prisma/nexus'

export const User = objectType({
  name: 'User',
  definition(t) {
    t.model.id()
    t.model.name()
    t.model.email()
    t.model.posts({
      pagination: false,
    })
  },
})
```

The above code doesn't work as it still provides the pagination fields in the `schema.graphql` file. But converting it to the object notation solves the issue

```ts
import { objectType } from '@prisma/nexus'

export const User = objectType({
  name: 'User',
  definition(t) {
    t.model.id()
    t.model.name()
    t.model.email()
    t.model.posts({
      pagination: {},
    })
  },
})
```

Also adding the boolean for the specific fields doesn't work

```ts
import { objectType } from '@prisma/nexus'

export const User = objectType({
  name: 'User',
  definition(t) {
    t.model.id()
    t.model.name()
    t.model.email()
    t.model.posts({
      pagination: { first: true, skip: false },
    })
  },
})
```

The above gives the following output in `schema.graphql`
```graphql
type User {
  email: String!
  id: ID!
  name: String
  posts(first: Int, skip: Int): [Post!]
}
```

### Library versions
prisma2: `2.0.0-preview-7`
@prisma/nexus: `0.0.1`
nexus: `0.12.0-beta.6`

Weakky:
Fixes #376

### DIFF
--- a/src/nexus-prisma/NexusPrismaBuilder.ts
+++ b/src/nexus-prisma/NexusPrismaBuilder.ts
@@ -284,9 +284,9 @@ export class NexusPrismaBuilder {
 
         args.push(...field.args.filter(a => paginationKeys.includes(a.name)))
       } else {
-        const paginationKeys = Object.keys(opts.pagination)
-
-        args.push(...field.args.filter(a => paginationKeys.includes(a.name)))
+        args.push(
+          ...field.args.filter(a => (opts.pagination as any)[a.name] === true),
+        )
       }
     }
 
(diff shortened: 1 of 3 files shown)


jasonkuhrt (approved):
LGTM

PR #405, opened 2019-09-12
feat: stop being a prisma generator

### LINKED ISSUE #395: BYONP: bring your own nexus-prisma (secondary: refactor)
1. **BYONP: Bring Your Own Nexus-Prisma**
I think it is confusing for the internal architecture and user-facing comprehension that `nexus-prisma` is a dependency of `prisma2` cli

2. **drop/rejig ncc build**
I think the ncc build step adds significant internal complexity, at least the way its setup now

### Motivation for BYONP:

1. From a user POV, conceptually if not actually/eventually, Prisma 2 has pluggable third party generators; thus generators are conceptually optional peer dependencies. We are making the system less coherent by not doing this in practice.

2. Users lose control over which version of nexus-prisma is in use. The release cycle of nexus-prisma is bound to prisma2 cli. This hurts users because all are sooner or later to want to make a match of `prisma2`/`nexus-prisma` that is not `latest` `latest`.

3. To a certain extent development on `nexus-prisma` is hurt because Namely/for example the local `example/` in this repo must do ugly hacks to make it use the local version of nexus-prisma rather than the one shipped by `prisma 2`.

4. To a certain extent it hurts docs because our local `example/` is somewhat compromised as an accurate reference due to point 3 above

### Motivation for no ncc build

1. When running the test suite, in debug mode over a test case, in VSCode, we cannot rely on breakpoints in the `src/` directory, since we are actually working with an `eval` statement in tests

2. Testing is made more complicated by requiring the build step prior to test run

3. TDD is compromised (maybe not just b/c of this but in part at least) because of the increased latency each test incurs

4. The project becomes harder to reason about. Layers upon layers of builds, generation, the generation doing its own generation (user time), ... it just isn't simple

---

I wouldn't be surprised if after further discussion we are able to turn the acc-build aspect of this issue into its own. That would be great.

I chalked this up to `type/feature` because of the BYONP aspect, but a large of this issue could be `type/refactor` too.

I am not yet deeply understanding of the benefits of the current approach (tradeoffs) CC @Weakky so did not mention them.

jasonkuhrt:
Will close #395
Blocks #406 

After some mental loops threatening infinite things came together and I’ve got `nexus-prisma` to stop being a prisma generator. This makes things _much_ simpler internally and I think also quite simpler for end-users. 

- No more ncc overhead
- No more [hacks caused by ncc](https://prisma-company.slack.com/archives/C8AKVD5HU/p1568253183003300)
- Easier integration tests
- Easier example management (proxy for: using dep graph, hello packing linking and all that good workflow stuff)
- User can control what version of nexus-prisma they want to use decoupled from their prisma2 version

#### TODO

- [x] [hide type gen in plugin construction](https://github.com/prisma/nexus-prisma/pull/405#discussion_r324574652)
- [x] document `Builder.build` for inline doc for users


example/src/index.ts
```diff
@@ -12,15 +12,17 @@ async function main() {
 
   await photon.connect()
 
-  const nexusPrisma = nexusPrismaPlugin({
+  const nexusPrisma = NexusPrisma.nexusPrismaPlugin({
     photon: ctx => ctx.photon,
```
jasonkuhrt: If needed user can pass custom photon path here

example/src/index.ts
```diff
@@ -12,15 +12,17 @@ async function main() {
...
   })
 
-  const schema: any = makeSchema({
+  NexusPrisma.generateTypes()
```
jasonkuhrt: Temporary until prisma/nexus#200

jasonkuhrt:
@timsuchanek asked:

> where will the generated typings live?

My answer:

> I think of the question as “where will nexus plugins put their typegen”. I think the answer for now is “wherever the core puts its typegen”. User can configure. Nexus actually currently forces users to do so. But sane defaults will come soon. The biggest point to make though is nexus plugin system will (my current vision) automate this and provide first class hooks for typegen contributions by plugins. This means the current additional overhead of nexus-prisma users needing to run typegen themselves (state of that pr) is temporary.

> Another lever available to us is nexus-framework but we can reach dx parity with prior nexus-prisma version re typegen without getting to that mystical layer. And must I’d say.

jasonkuhrt:
more analysis in notion https://www.notion.so/prismaio/nexus-prisma-as-Prisma-Generator-or-Nexus-Generator-18fff59c716c4c5fbd129c22c3b663d1

example/src/index.ts
Weakky: Why can't we generate the types when calling `nexusPrismaPlugin`, just like `nexus` generate types when calling `makeSchema` when not in production?

src/builder.ts
```diff
@@ -27,31 +26,37 @@ interface FieldPublisherConfig {
...
-    }
-
-    this.dmmf = new DMMFClass(transformedDMMF as any)
+  constructor(protected options: Options) {
+    // TODO Default should be updated once resolved:
+    // https://github.com/prisma/photonjs/issues/88
+    // TODO when photon not found log hints of what to do for the user
+    // TODO DRY this with same logic in typegen
+    const photonPath = options.photonPath || '@generated/photon'
+    const transformedDMMF = transformDMMF(require(photonPath).dmmf)
```
Weakky: Once we remove the exported dmmf from photon, this step will be too expensive to be run at runtime and therefore to be viable in production, especially in lambda's
That's one of the reason why nexus-prisma is currently a generated library: to remove the performance overhead of computing DMMF

jasonkuhrt: Discussion between @Weakky and I since this comment was written contributed to 

https://www.notion.so/prismaio/Nexus-prisma-as-a-regular-library-178b5232def24800ad4c7e17fcc1201b

Conclusion was that we can rely on dmmf via photon or via its an import in its own right, either way though, the key, an import.

example/src/index.ts
jasonkuhrt: Done!

jasonkuhrt:
@Weakky finished another pass and ready for your review 2. Now that we've all aligned on the macro, lets move on to the micro :D

src/builder.ts
```diff
@@ -27,31 +28,74 @@ interface FieldPublisherConfig {
   ordering?: boolean | Record<string, boolean>
 }
 
+export interface Options {
+  photon?: (ctx: any) => any
+  shouldGenerateArtifacts?: boolean
+  inputs?: {
+    photon?: string
```
Weakky, line 34: I didn't pay attention to this addition (`inputs.photon`), but I'd say that sucks 😞
While it's a simple field, we're degrading the surface API to reduce our internal complexity and I don't like it

Not only are we leaking technical details to the surface API (`inputs.photon` is the path we need to require DMMF and users don't need to know about that), but we also end up with a weird duplicated API: `photon` and `inputs.photon`, which IMO doesn't make sense from a user perspective.

We could for sure enhance it by renaming `inputs.photon` to something like `inputs.photonPath`, but it'd still force users to input photon's path twice (once to import photon's instance to pass it to their graphql server context, and a second time to pass it to nexus-prisma) on the worst case scenario, and again, I think that's not very elegant

I'd suggest waiting before we make a decision on the DMMF generator. If we agree on the DMMF generator, we'd end up with a predictable location to require the DMMF and therefore remove the need for `inputs.photon`

jasonkuhrt: Agree it sucks, naming was an attempt to use symmetrical terms (i/o) and following from nexus api (outputs).

This complexity follows from the lack of guarantees from generators that they will play nice with package/dep system. I think we should view the complexity here as a symptom of an upstream problem. Related is https://github.com/prisma/specs/issues/152 fwiw...

I do think its an ok tradeoff if it means they don't have to write a generator config block. There's a clear tradeoff there. And it seems reasonable 🤔 .

Issue #400, opened 2019-09-09
Integration tests over typegen


jasonkuhrt:
Given an app fixture:

   - schema.prisma
   - graphql.ts

Run an integration test:

  1. run photon generation
  2. run nexus generation
  3. snapshot typegen
  4. snapshot graphql schema
  5. assert app Type check
  6. assert (manually) typegen snap
  7. assert (manually) graphql schema snap

Options:

  - If we commit the typegen we can rely on on VSCode TS checks while editing the fixture
  - It follows, from above, that if we commit the typegen, we don't really need to snapshot it, because committing to the repo is already achieving that (albeit a workflow of `git diff`)

Open:

* Should we have one big or many smaller integration tests? I think one big one is a lot easier to manage. But it may frustrate attempts to capture particular edge-cases or regressions. Maybe a general big one that tries to exercise basically all features, with a bunch of tiny ones for specific issues. I think that could work. What we should avoid is a bunch of overlapping medium sized ones, confusing what tests live where and where new tests should go.

jasonkuhrt:
@Weakky Please review my proposed plan, makes sense?

jasonkuhrt:
While we wait to settle https://github.com/prisma/specs/issues/131 I tried to leverage [yarn `resolutions`](https://stackoverflow.com/questions/40226639/how-do-i-override-nested-dependencies-with-yarn) feature to make `prisma2 generate` use local `nexus-prisma` version. But discovered that's not possible because `prisma2` [is published with an ncc step](https://prisma-company.slack.com/archives/C8AKVD5HU/p1568227459038400?thread_ts=1568224651.033700&cid=C8AKVD5HU).

PR #406, opened 2019-09-12
tests: setup integration test system

jasonkuhrt:
This PR does not attempt to actually write out integration tests. This PR is already large. It aims to setup the system to conduct integration tests.

Contributes toward #400


jasonkuhrt:
Just a meta failure now. Due to the customer matcher we have registered not being picked up by jest.

tests/__setup.ts
```diff
@@ -0,0 +1,66 @@
...
+    )
+    .join('\n')}`
+
+  const jestReport = `${summaryReport}\n\n${tsReport}`
+
+  return jestReport
+}
+
+expect.extend({
+  toTypeCheck(tsConfigFilePath: string) {
```
jasonkuhrt: I built similar [for nexus](https://github.com/prisma/nexus/blob/develop/tests/_setup.ts). Inline for now but I could see/I think we should extract this out into a jest-ts-test-utils package.

jasonkuhrt:
Failure on ci that I cannot reproduce locally, doing the exact same steps, seemingly. Need to investigate more deeply.

package.json
```diff
@@ -24,16 +24,20 @@
...
   "devDependencies": {
     "@prisma/photon": "^0.2.6",
+    "@types/fs-extra": "^8.0.0",
     "@types/graphql": "^14.2.3",
     "@types/jest": "^24.0.18",
     "@types/node": "^12.7.2",
     "@types/pluralize": "^0.0.29",
     "@types/tmp": "^0.1.0",
+    "fs-extra": "^8.1.0",
+    "graphql": "^14.5.4",
```
jasonkuhrt, line 32: To make peer dep warning from nexus go away

src/builder.ts
```diff
@@ -516,11 +515,16 @@ export class NexusPrismaBuilder {
...
+    // typegen should contribute (e.g. via interface merging) to it
+    // the scalars found in DMMF.
+    if (dedupScalarNames.includes('DateTime' as any)) {
       scalars.push(dateTimeScalar)
     }
-
-    if (dedupScalarNames.includes('UUID')) {
+    if (dedupScalarNames.includes('UUID' as any)) {
       scalars.push(uuidScalar)
     }
```
jasonkuhrt, line 545: Actual bug finds!

Weakky: I think the reason why it used to be a string is because of the problem you're mentioning in the comment. Even if nexus includes custom scalars in the `AllOutputTypes` type, the type system won't be able to catch them at "dev time" anyway as there's no nexus generated typegen file yet.
In the end, IMO, what matters here is that the surface/public API remain strongly typed. Internally though, we cannot properly type everything as some parts rely on the generated typegen file which happens at runtime. For these cases, I think it's not too bad to use "general types" (such as `string` in this case) as I don't see any other viable option

tests/app.test.ts
```diff
@@ -0,0 +1,105 @@
...
+// - show we gracefully handle case of photon import failing
+
+it('integrates together', async () => {
+  // Setup file system vars & helpers
+  //
+  const projectRoot = path.join(__dirname, '/__app')
+
+  async function projectReadFile(relPath: string): Promise<string> {
+    return fs.readFile(path.join(projectRoot, relPath)).then(b => b.toString())
+  }
```
Weakky: `async` is not useful if you're using promise-style callbacks in the function body

Weakky (changes requested):
The integration setup + test looks all good to me, nice work 🙏

tests/app.test.ts
jasonkuhrt: It removes a vsc warning. Have a better way to do that?

src/builder.ts
jasonkuhrt: > the type system won't be able to catch them at "dev time" anyway as there's no nexus generated typegen file yet.

Do you mean dev time of nexus-prisma?

> In the end, IMO, what matters here is that the surface/public API remain strongly typed.

agree

> Internally though, we cannot properly type everything as some parts rely on the generated typegen file which happens at runtime.

My conclusion was different. We can import types from `nexus` and correctly type our plugin.

tests/app.test.ts
jasonkuhrt: image

Sigh, https://github.com/Microsoft/vscode/issues/61326

Going to disable my vscode `suggestionActions` ✌️

src/builder.ts
jasonkuhrt: @Weakky let's pick up the topic of thread next week.

PR #408, opened 2019-09-17
Fix nested inputs aliasing

### LINKED ISSUE #331: Nested filter on related model not working
(Prisma2/nexus)

I am using the following (simplified) model:
```graphql
model Staff {
  id         String       @default(cuid()) @id @unique
  user       User?
  active     Boolean?
}
model User {
  id         String        @default(cuid()) @id @unique
  email      String        @unique
  staff      Staff?
}
```
Using photon, the following query works:
```ts
  const staff = await photon.users.findMany({
    where: {
      staff: { active: true },
    },
  });
```
However, the following graphql query, exposed through nexus-prisma, fails:
```graphql
query {
  findManyUser(
    where: { staff: { active: true } }
  ) {
    id
    email
  }
}
```
with the error message:
```
Field "active" is not defined by type QueryFindManyUserWhereInput.
```
The nexus-prisma type generation seems to be incorrect, `active` should not be of type `QueryFindManyUserWhereInput`, I think.

Weakky:
Fixes #331

### DIFF
--- a/src/nexus-prisma/NexusPrismaBuilder.ts
+++ b/src/nexus-prisma/NexusPrismaBuilder.ts
@@ -9,6 +9,7 @@ import {
   flatMap,
   getCRUDFieldName,
   nexusOpts as nexusFieldOpts,
+  partition,
 } from '../utils'
 import {
   defaultArgsNamingStrategy,
@@ -27,10 +28,14 @@ interface FieldPublisherConfig {
   ordering?: boolean | Record<string, boolean>
 }
 
+interface CustomInputArg {
+  arg: DMMF.SchemaArg
+  type: DMMF.InputType | DMMF.Enum | 'scalar'
+}
+
 export class NexusPrismaBuilder {
   protected readonly dmmf: DMMFClass
   protected visitedInputTypesMap: Record<string, boolean>
-  protected whitelistMap: Record<string, string[]>
   protected argsNamingStrategy: IArgsNamingStrategy
   protected fieldNamingStrategy: IFieldNamingStrategy
 
@@ -49,7 +54,7 @@ export class NexusPrismaBuilder {
     this.argsNamingStrategy = defaultArgsNamingStrategy
     this.fieldNamingStrategy = defaultFieldNamingStrategy
     this.visitedInputTypesMap = {}
-    this.whitelistMap = {}
+
     if (!this.params.photon) {
       this.params.photon = ctx => ctx.photon
     }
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -27,7 +42,7 @@ export function flatMap<T, U>(
 }
 
 export function nexusOpts(param: {
-  type: string
+  type: core.AllInputTypes | core.AllNexusInputTypeDefs<string>
```
jasonkuhrt: What's the difference between these two types?

tests/schema.test.ts
```diff
@@ -289,3 +289,46 @@ test("it does not expose pagination", async () => {
 
   expect(printSchema(schema)).toMatchSnapshot()
 })
+
+test('alias top-level input types only if not customized', async () => {
```
jasonkuhrt:
```suggestion
test('renames top-level input types that are customized', async () => {
```

Or do I misunderstand?

tests/__snapshots__/schema.test.ts.snap
```diff
@@ -1,5 +1,78 @@
...
+input PostWhereInput {
+  id: StringFilter
+  AND: [PostWhereInput!]
+  OR: [PostWhereInput!]
+  NOT: [PostWhereInput!]
+  author: UserWhereInput
+}
+
+type Query {
+  posts(where: QueryFindManyPostWhereInput, skip: Int, after: String, before: String, first: Int, last: Int): [Post!]
```
jasonkuhrt: Aside: strikes me that this should be named `QueryPostsWhereInput`

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -262,59 +266,69 @@ export class NexusPrismaBuilder {
...
+    const type = this.dmmf.getInputType(inputTypeName)
+
+    // Do not alias type name if type is not customized
+    if (input === true) {
+      return type
+    }
+
+    const argsNames = Object.keys(input)
+
+    // Alias type name if some fields are whitelisted
```
jasonkuhrt:
```suggestion
    // Rename the InputObject type definition if some fields are whitelisted
```

Feel like `alias` is not the word we want 🤔

src/nexus-prisma/NexusPrismaBuilder.ts
```diff
@@ -229,31 +235,29 @@ export class NexusPrismaBuilder {
...
-            whereArg.inputType.type,
-            whereArg.inputType.kind,
-          )
-        ] = Object.keys(opts.filtering).filter(
-          fieldName => (opts.filtering as any)[fieldName] === true,
+          `Could not find filtering argument in ${prismaModelName}.${
+            field.name
+          }(${field.args
+            .map(a => `${a.name}: ${a.inputType.type}`)
+            .join(', ')}): ${field.outputType.type}`,
```
jasonkuhrt: Bit hard to read this

jasonkuhrt (approved):

src/nexus-prisma/NexusPrismaBuilder.ts
Weakky: Agreed but not sure how to factor it better. It generates something like `Query.posts(first: Int, last: Int, before: String, after: String)`

Weakky: I reverted it anyway, we'll think about better errors in another pr

PR #409, opened 2019-09-17
tests: migrate partial integration tests to full

### LINKED ISSUE #400: Integration tests over typegen

jasonkuhrt:
closes #400


Weakky:
I'm worried it might get harder and harder to identify an issue after breaking something or to make sure the voluntary changes we made are properly reflected in the snapshot

jasonkuhrt:
> I'm worried it might get harder and harder to identify an issue after breaking something or to make sure the voluntary changes we made are properly reflected in the snapshot

I share that concern too. I think though we can stand to benefit from putting a small/medium app under test. This is definitely a testing experiment.

Also, I think we'll soon be ready to start introducing some unit tests. This would help us with error pinpointing. Integration tests would be more to give us system-level confidence.

What do you think?

jasonkuhrt:
Let's see how it goes :)

PR #410, opened 2019-09-19
fix: enums as output type should be published too

Weakky:
## Description

- BREAKING: A new `types` property was added to `nexusPrismaPlugin`
- Fix empty enums filters type
- Refactor types publishing system (into a `Publisher` class)
- Enums are now published even when referenced as output type
- Any user-defined type won't be re-published by nexus-prisma (input type, enum, scalar..)
- Added tests


jasonkuhrt:
Cleaned up the false negative test failure. Now dealing with a remaining actual type error that occurs when the typegen is present.

PR #412, opened 2019-09-20
feat(gql-types): drop input suffix naming

jasonkuhrt:
I would like us to consider dropping the `Input` suffix on GQL input types:

* makes already long names even longer
* so, makes schema feel more heavy and less readable
* redundant in the sense that the underlying type is already called `Input`

The implementation is whatever right now, not intended to be considered. Just the idea. Consider the effect on [the GQL Schema](https://github.com/prisma/nexus-prisma/pull/412/files#diff-0611fb3c878b9df2656def8166d0cc6f).


jasonkuhrt:
Things we can try re moving the discussion forward:

- gather more research from the community e.g. @Weakky mentioned pulling some [Apollo examples](https://blog.apollographql.com/designing-graphql-mutations-e09de826ed97)
- Ask our external slack community
- Ask internally (this design is inherited from GraphCool so lets understand how it came to be originally)
- Given we now have https://github.com/prisma/nexus-prisma/issues/309 maybe we can consider how this is a userland decision. Would still then need to decide if its opt-in or opt-out.

jasonkuhrt:
After rebase things are broken, need to redo some amount of work. I believe my work was done before the publisher refactor.

PR #418, opened 2019-09-21
fix: convert \ to / on windows for photon import

wardpeet:
When running `prisma2 generate` on windows I get `import * as photon from '@generated\photon';` inside `nexus-prisma/index.d.ts`. This PR converts \ into /.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,10 +9,10 @@
     "dist"
   ],
   "scripts": {
-    "clean": "del-cli -f dist",
+    "clean": "rm -rf dist",
     "build": "yarn -s clean && tsc",
     "dev": "tsc --watch",
-    "prepublish": "yarn -s build",
+    "prepublishOnly": "yarn -s build",
     "style": "prettier --write src/**/*.ts test/**/*.ts example/**/*.ts",
     "test": "jest"
   },
@@ -29,7 +29,6 @@
     "@types/node": "^12.7.2",
     "@types/pluralize": "^0.0.29",
     "@types/tmp": "^0.1.0",
-    "del-cli": "^3.0.0",
     "graphql": "^14.5.4",
     "jest": "^24.9.0",
     "jest-watch-typeahead": "^0.4.0",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "dist"
   ],
   "scripts": {
-    "clean": "rm -rf dist",
+    "clean": "del-cli -f dist",
     "build": "yarn -s clean && tsc",
     "dev": "tsc --watch",
     "prepublish": "yarn -s build",
@@ -29,6 +29,7 @@
     "@types/node": "^12.7.2",
     "@types/pluralize": "^0.0.29",
     "@types/tmp": "^0.1.0",
+    "del-cli": "^3.0.0",
     "graphql": "^14.5.4",
     "jest": "^24.9.0",
     "jest-watch-typeahead": "^0.4.0",
--- a/src/typegen.ts
+++ b/src/typegen.ts
@@ -37,8 +37,9 @@ export function doGenerate(
 }
 
 export function render(dmmf: DMMF.DMMF, photonPath: string) {
+  // photonPath contains \\ in windows. We'll replace it with a / if it occurs.
   return `\
-import * as photon from '${photonPath}';
+import * as photon from '${photonPath.replace(/\\g/, '/')}';
 import { core } from 'nexus';
 // Types helpers
 ${renderStaticTypes()}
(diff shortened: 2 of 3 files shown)


src/typegen.ts
```diff
@@ -37,8 +37,9 @@ export function doGenerate(
 }
 
 export function render(dmmf: DMMF.DMMF, photonPath: string) {
+  // photonPath contains \\ in windows. We'll replace it with a / if it occurs.
   return `\
-import * as photon from '${photonPath}';
+import * as photon from '${photonPath.replace(/\\g/, '/')}';
```
janpio, line 42: Hm, that probably does work - but we should probably try to fix the actual `photonPath` to be correct in the first place, right? Who is calling `generate` and where do the options of that come from?

wardpeet: that's probably a good idea but I had no idea where it came from ^^ so any pointers would be great ^^

package.json
```diff
@@ -9,7 +9,7 @@
     "dist"
   ],
   "scripts": {
-    "clean": "rm -rf dist",
+    "clean": "del-cli -f dist",
```
janpio: this unrelated change is useful in any case for windows compat

wardpeet: If necessary I can create a separate PR.

src/typegen.ts
janpio: I am pretty sure @Weakky or @jasonkuhrt can give us a pointer where we should look <3

package.json
jasonkuhrt: Hey @wardpeet did you hit this in user land or are you trying to make the project more friendly to develop on Widows machines?

wardpeet: @jasonkuhrt I'm a windows user so I've hit this issue when doing a `npm install`. If you want to not trigger it on `npm install` you have to use `prepublishOnly` as described in https://docs.npmjs.com/misc/scripts

src/typegen.ts
jasonkuhrt: This value comes from what users pass into their plugin config. And there's a default if nothing is passed: https://github.com/prisma-labs/nexus-prisma/blob/61d6efa56301d7d4419ef9e0747447c0801b7698/src/builder.ts#L99.

I can see how the use of hardcoded `/` might not be portable. However this PR converts from `\` to `/` anyways, so I'm not fully understanding.

@wardpeet could you explain the issue you're facing a bit more?

package.json
jasonkuhrt: @wardpeet Thanks, remember now. Yep what we want is `prepublishOnly`. Could you make that change in this PR~, and~ undo this bit? I'll make the script fix now on `next`.

src/typegen.ts
wardpeet: Definitely, when running `prisma2 generate` on windows. The `src/generated/nexus` file is being generated. 

On windows, I get this content:
```
/**
 * This file was automatically generated by GraphQL Nexus
 * Do not make changes to this file directly
 */

import * as photon from "@generated\photon"



declare global {
  interface NexusGenCustomOutputProperties<TypeName extends string> {
    crud: NexusPrisma<TypeName, 'crud'>
    model: NexusPrisma<TypeName, 'model'>
  }
}
```

I'm not sure how `@generated\photon` is passed into prisma?

In my index.ts file I have the following config
```
const schema = makeSchema({
  types: [allTypes, nexusPrisma],
  outputs: {
    schema: path.join(__dirname, './generated/schema.graphql'),
    typegen: path.join(__dirname, './generated/nexus'),
  },
  typegenAutoConfig: {
    // debug: true,
    sources: [
      {
        source: '@generated/photon',
        alias: 'photon',
      },
      {
        source: path.join(__dirname, 'types.ts'),
        alias: 'ctx',
      },
    ],
  },
});
```

To be honest I have no idea what i'm doing, i'm just copy pasting from prisma website :joy:

jasonkuhrt: My last comment was off topic b/c was I was talking about the wrong thing.

For photon path the default is here https://github.com/prisma/nexus-prisma/blob/8c0164876bba33c459fa044119487fc50331cb52/src/builder.ts#L92 and again its something that the user can pass to the nexus-prisma plugin config.

@wardpeet IIUC you need `@generated\photon` to be `@generated/photon` in the typegen?

wardpeet: I need `@generated/photon` as \ is an escape char so it won't resolve the `@generated/photon` typings.

I'm not overriding the config.input.photon also it looks like `prisma2 generate` is the culprit as it creates the wrong import inside `@generated/nexus-prisma/index.d.ts`
image

jasonkuhrt: Also the next version of nexus-prisma is not a prisma generator, so this issue can be considered already fixed. A release just has to be made. I will therefore close this PR. Release will be made today or tomorrow.

wardpeet: ok sounds good 👌

PR #441, opened 2019-10-04
fix(publishing): publish dmmf objects

### LINKED ISSUE #440: dmmf object BatchPayload is never published
First reported by @mikeslade in https://github.com/prisma-labs/nexus-prisma/issues/383#issuecomment-538073790.

Verified this issue by trying to add a `*Many*` mutation to the integration app test.

jasonkuhrt:
Will close #440

### DIFF
--- a/src/publisher.ts
+++ b/src/publisher.ts
@@ -1,6 +1,6 @@
 import * as Nexus from 'nexus'
 import * as DMMF from './dmmf'
-import { nexusFieldOpts, partition } from './utils'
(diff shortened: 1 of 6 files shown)


Weakky:
With that fix, it means that one could just expose the `Query` type and have everything else published automatically.

This is more or less what we'd like to have with the prototyping mode, but we specifically wanted to forbid that in production, because that might lead into unintended published fields (such as user's password)

PR #443, opened 2019-10-05
fix(typegen): enum-field publishing on models

### LINKED ISSUE #442: Support enum model methods in typegen
This was raised in #383 by users saying the issue was not actually resolved. It turns out that there was a different issue resulting from typegen not supporting model methods returning enum  (aka. fields of kind enum). The actual runtime worked just fine (e.g. if you were a JS user). Thanks to @mikeslade et al. for raising this clear and loud!

jasonkuhrt:
Will close #442


Weakky:
Great job reverse-engineering the terrible typegen 🙈

PR #451, opened 2019-10-07
fix: publish object types that don't map to prisma models…

Weakky:
Object types should be auto-published only in case their names don't match a Prisma model name.

This fix allows us to properly auto-publish `BatchPayload`, but keep forcing users to define their top-level object types

Related to: https://github.com/prisma-labs/nexus-prisma/pull/441#issuecomment-538970758

### DIFF
--- a/tests/outputs.test.ts
+++ b/tests/outputs.test.ts
@@ -1,0 +1,31 @@
+import * as Nexus from 'nexus'
+import { generateSchemaAndTypesWithoutThrowing } from './__utils'
+
+it('only publishes output types that do not map to prisma models', async () => {
+  const datamodel = `
+  model User {
+    id  Int @id
+    name String
+  }
+`
+  const Query = Nexus.objectType({
+    name: 'Query',
+    definition(t: any) {
+      t.crud.user()
+    },
+  })
+
+  const Mutation = Nexus.objectType({
+    name: 'Mutation',
+    definition(t: any) {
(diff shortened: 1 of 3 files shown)


tests/outputs.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    },
+  })
+
+  const {
+    schema,
+    typegen,
+    missingTypes,
+  } = await generateSchemaAndTypesWithoutThrowing(datamodel, [Query, Mutation])
+
+  expect(Object.keys(missingTypes)).toContain('User')
```
jasonkuhrt: Can we strengthen this to a "same members" check (order does not matter). It should be an error if `BatchPayload` shows up here for example.

tests/outputs.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  expect(schema).toMatchSnapshot('schema')
+  expect(typegen).toMatchSnapshot('typegen')
```
jasonkuhrt: I wonder if we need snapshots for these non-integration tests.

jasonkuhrt (approved):

tests/outputs.test.ts
Weakky: Removed the snapshots, you're right. I added the snapshot because I wanted to make sure that the missing type was properly replaced in the schema, but it's out of scope for this test

PR #452, opened 2019-10-07
fix: only publish non-null list + list members

### LINKED ISSUE #437: Prisma2: crud findMany method returns optional value
Given Nexus config

```typescript
export const Query = extendType({
  type: 'Query',
  definition(t) {
    t.crud.merchants()
  },
})
```

The following schema is generated

```
type Query {
  merchants(after: String, before: String, first: Int, last: Int, skip: Int): [Merchant!]
}
```

It seems that correct return type should be `[Merchant!]!` instead.

jasonkuhrt:
Will close #437


jasonkuhrt:
Nice [upstream fix](https://github.com/prisma/prisma-engine/pull/52) @sameoldmadness! @timsuchanek 👋I suspect my code comment in this PR is not accurate. Can you confirm/deny? It does seem that way though upon cursory glance at the photon type definitions, example:

image

Weakky (changes requested):
I suspect you forgot to add the same logic to the `buildModel()` function. There can be model fields that returns lists as well. eg: `type Blog { posts(...): [Post!]! }`

Maybe it's worth refactoring the `listNullableSettings` logic into its own function and apply it on both `buildCrud` and `buildModel`

Weakky (approved):
Changes approved 🙌

Re the comment

image

I'm not sure to understand what you mean here. Specifically, the diff between `lists` and `lists-of-values` 🤔

PR #459, opened 2019-10-10
docs: intro, concepts, reference

### LINKED ISSUE #186: Docs with Prisma 2 aren't clear. Throws error.
I am trying https://nexus.js.org/docs/database-access-with-prisma-v2 but the docs aren't clear & throws error.

For example, before `1.3. Add dependencies`, add a `cd myblog/` step which is basic I know but for beginners, it is helpful.

Then Idk why there is `npm init -y` when we already have `package.json` created by `prisma2 init myblog` which I think should be removed as well.

Again `tsconfig.json` is already created by the `prisma2 init myblog` command.

`1.5. Configure start script in package.json` gives another error if I run `yarn start`

Then in `2. Define models & generators` they say edit `project.prisma` but there is no such file. There exists `schema.prisma` in `prisma/` folder. If that is the file we have to edit, then I guess it's already edited.

Let's assume I edited it & run `3. Migrate database`. In that, I run `prisma2 lift save --name="init"` which gives `Everything up-to-date` but running `prisma2 lift up` gives error.

> Error: Error in lift engine for rpc migrationProgress:
  thread 'main' panicked at 'Could not load migration from database. Migration name was: 20190624162221-init', src/libcore/option.rs:1034:5

It basically is terrible at this point. I don't even know if I'm doing it correctly. It throws a ton of errors. Docs has a lot of problem. Would love if it's been edited properly so it works fine.

Unfortunately, I haven't made it work so can't send a PR.

### LINKED ISSUE #447: Getting started flow is missing from README

### LINKED ISSUE #489: reference docs
We're currently lacking exhaustive API and general reference docs.

jasonkuhrt:
[preview](https://github.com/prisma-labs/nexus-prisma/tree/docs/revamp)

Will close #489
Will close #447 
Will close https://github.com/prisma-labs/nexus/issues/186


schickling (approved):
Looks great!

README.md
```diff
@@ -5,11 +5,1119 @@
...
+`nexus-prisma` is a plugin for bridging [Prisma](https://www.prisma.io) and [Nexus](https://nexus.js.org). It extends the Nexus DSL `t` with `.model` and `.crud` making it easy to expose Prisma models and operations against them in your GraphQL API. The resolvers for these operations (pagination, filtering, ordering, and more), are dynamically created for you removing the need for traditional ORMs/query builders like TypeORM, Sequelize, or Knex. And when you do need to drop down into custom resolvers a [Photon](https://photonjs.prisma.io) instance on `ctx` will be ready to serve you, the same great tool `nexus-prisma` itself bulids upon.
 
-## Examples
+**Table of Contents**
```
nikolasburk: I feel like this could be an `H1` heading as well

nikolasburk: (ah or does this break the auto-generated TOC because it would include itself?)

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+```
+npm install nexus-prisma
+```
+
+## Example
+
+Given a Prisma schema like:
+
+```prisma
```
nikolasburk, line 73: I like your idea of making a sample schema more fun. I suggest that we try to come up with a canonical schema that we can use as foundation for _all_ our content, thinking about what features we want to show. _For now_ I suggest to go with the `User` and `Post` models we're using for all other examples as well (maybe reduce the amount of fields of the models to make it a bit simpler).

README.md
```diff
@@ -5,11 +5,1119 @@
...
+}
+```
+
+</details>
+
+<br>
+
+You can find a runnable version of this and other examples at [prisma-labs/nexus-examples](/TODO).
+
+## Reference
```
nikolasburk: Maybe call this `API Reference`?

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+### `t.model`
+
+`t.model` is for projecting your models’ fields onto your `Object`s. It is only available inside `Object` definitions. `t.model` properties are called "Field Projectors" and there is one per field on your respective model. They accept configuration allowing you to control how the model's fields are seen through your GraphQL schema.
```
nikolasburk: I'm a bit confused by the inline code formatting for `Object`? are you referring to `objectType` here? I think for inline code formatting, we should try to apply it only to concepts that are literally code pieces, so if `Object` stands for `objectType` here, maybe indeed use `objectType` instead?

README.md
```diff
@@ -5,11 +5,1119 @@
...
+`nexus-prisma` is a plugin for bridging [Prisma](https://www.prisma.io) and [Nexus](https://nexus.js.org). It extends the Nexus DSL `t` with `.model` and `.crud` making it easy to expose Prisma models and operations against them in your GraphQL API. The resolvers for these operations (pagination, filtering, ordering, and more), are dynamically created for you removing the need for traditional ORMs/query builders like TypeORM, Sequelize, or Knex. And when you do need to drop down into custom resolvers a [Photon](https://photonjs.prisma.io) instance on `ctx` will be ready to serve you, the same great tool `nexus-prisma` itself bulids upon.
```
nikolasburk: I suggest to always refer to Nexus as "GraphQL Nexus" in public communication. Many people won't really know what "Nexus" is when they're reading our content for the first time, so I suggest to always call out the association with GraphQL explicitly.
image

nikolasburk: Let's refer to "Photon" as "Photon.js" to be more precise (or use inline code for `Photon`)

README.md
```diff
@@ -5,11 +5,1119 @@
...
+}
+
+model User {
+  id     String @id @default(cuid())
+  handle String
+  posts  Post[]
+}
+
+model Post {
+  id     Int    @id
```
nikolasburk: for consistency, should we make the `id` of type `String @id @default(cuid())` as well?

README.md
```diff
@@ -5,11 +5,1119 @@
...
+import { makeSchema } from 'nexus'
+import * as types from './types'
+
+export const schema = makeSchema({
+  types: [...types, createNexusPlugin({ types })],
+  outputs: true,
+})
+```
+
+Generate your Photon client:
```
nikolasburk: Let's refer to it as "Photon.js database client" in written content to be hyper explicit.

README.md
nikolasburk: > `t.model` is for projecting your models’ fields onto your `Object`s.

I think _I_ understand this sentence because I have some context of Prisma and Nexus but I imagine it to be difficult to grasp for a newcomer since it references Prisma and Nexus concepts in one sentence. I suggest to be a bit more explicit, something like: "`t.model` is for projecting the fields of your [Prisma models](https://github.com/prisma/prisma2/blob/master/docs/data-modeling.md#models) onto your GraphQL types which are defined using the [`objectType`](https://nexus.js.org/docs/api-objecttype) function from the GraphQL Nexus API."

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+## Installation
+
+```
+npm install nexus-prisma
+```
+
+## Example
+
+Given a Prisma schema like:
```
nikolasburk: Link "Prisma schema" to https://github.com/prisma/prisma2/blob/master/docs/prisma-schema-file.md

README.md
nikolasburk: > `t.model` properties are called "Field Projectors"

Question: Are we sure we want to uppercase concepts like field projectors? To me these are regular features (not _names_) so I'd rather lowercase them. I've been in that situation a number of times when documenting Prisma features, and I feel like defaulting to lowercasing generally feels more natural. Otherwise it's hard to draw a boundary when something gets lowercased and when it gets uppercased.

Second note: I'd write this in _italics_ instead of "quotes". IMO quotes have specific semantics (imagine someone doing air quotes on this while reading the sentence), but here you just want to emphasize _field projectors_ as a new concept, so I think italics (or maybe bold) are better suited.

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+## Reference
+
+### `t.model`
+
+`t.model` is for projecting your models’ fields onto your `Object`s. It is only available inside `Object` definitions. `t.model` properties are called "Field Projectors" and there is one per field on your respective model. They accept configuration allowing you to control how the model's fields are seen through your GraphQL schema.
+
+#### Field Projectors
+
+`t.model` will have Field Projectors for the model whose name matches that of the `Object`. If the `Object` is of a name that does not match any of your models then `t.model` becomes a function allowing you to specify the mapping.
```
nikolasburk: > `t.model` will have Field Projectors for the model whose name matches that of the `Object`.

Again, I suggest to be a bit more explicit and call out that model is a Prisma concept: "`t.model` will have Field Projectors for the Prisma model whose name matches that of the `Object`" 

Second similar note as before: should we refer to `Object` as `objectType` to be more precise?

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+objectType({
+  name: 'User',
+  definition(t) {
+    t.model.name()
+  },
+})
+```
+
+```gql
```
nikolasburk, line 363: I'd add a quick sentence to provide some context before the `gql` code block: "This corresponds to the following type definition in SDL" (or something similar)

README.md
```diff
@@ -5,11 +5,1119 @@
...
+type User {
+  name: String
+}
+```
+
+```ts
+// Project User model fields `id` & `name` onto Person Object
```
nikolasburk: the example doesn't show an `id` field?

README.md
```diff
@@ -5,11 +5,1119 @@
...
+### `t.model`
+
+`t.model` is for projecting your models’ fields onto your `Object`s. It is only available inside `Object` definitions. `t.model` properties are called "Field Projectors" and there is one per field on your respective model. They accept configuration allowing you to control how the model's fields are seen through your GraphQL schema.
+
+#### Field Projectors
+
+`t.model` will have Field Projectors for the model whose name matches that of the `Object`. If the `Object` is of a name that does not match any of your models then `t.model` becomes a function allowing you to specify the mapping.
+
+```ts
+// Project User model fields `id` & `name` onto User Object
```
nikolasburk: the example doesn't show an `id` field?

README.md
nikolasburk: > and there is one per field on your respective model.

suggested simplification: "and there is exactly one per field on each model."

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+objectType({
+  name: 'Person',
+  definition(t) {
+    t.model('User').name()
+  },
+})
+```
+
+```gql
```
nikolasburk, line 417: one thing to note for these examples, I think it would be helpful to know what Prisma model these are based on. you could e.g. add a note to the beginning of the section saying that these examples are based on the following Prisma model:

```
model User {
  id     String @id @default(cuid())
  name   String
}
```

README.md
```diff
@@ -5,11 +5,1119 @@
...
+```ts
+objectType({
+  name: 'User',
+  definition(t) {
+    t.model.name({ alias: 'handle' })
+  },
+})
+```
+
+```gql
```
nikolasburk, line 512: I'd add a quick sentence to provide some context before the `gql` code block: "This corresponds to the following type definition in SDL" (or something similar)

README.md
```diff
@@ -5,11 +5,1119 @@
...
+}
+```
+
+#### `ordering` `pagination` `filtering` Options
+
+Only available for list type model fields. Please refer to TODO for details.
+
+### `t.crud`
+
+`t.crud` is for exposing operations against your projected models. It is only available inside `Query` and `Mutation` definitions. `t.crud` properties are called "Operation Publishers". They accept configuration allowing you to control how the operation is seen through your GraphQL schema.
```
nikolasburk: I suggest to be a bit more explicit again: "t.crud` is for exposing CRUD operations against your projected [Prisma models](https://github.com/prisma/prisma2/blob/master/docs/data-modeling.md#models)."

nikolasburk: > "Operation Publishers"

same questions/notes as with "Field Projectors" (suggestion to lowercase and spell in italics)

nikolasburk: > It is only available inside `Query` and `Mutation` definitions.

maybe again be explicit that `Query` and `Mutation` are Nexus concepts, e.g. with: It is only available inside the `Query` and `Mutation` types that are defined with the GraphQL Nexus API.

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+`t.crud` reflects a subset of [Photon](https://photonjs.prisma.io)'s capabilities. For every one model in your Prisma schema there are ten corresponding Operation Publishers.
```
nikolasburk: Photon -> Photon.js

nikolasburk: > For every one model in your Prisma schema there are ten corresponding Operation Publishers.
 
For some reason reading the "ten" like this feels a bit weird. I suggest to emphasize that this is consciously written like this by adding "exactly": " there are exactly ten"

README.md
```diff
@@ -5,11 +5,1119 @@
...
+<example>
+  <GraphQL client operations>
+  <GraphQL schema>
+  <Nexus/Nexus-Prisma type defs>
+  <Prisma schema model defs>
+```
+
+#### One-Create Operation
+
+Allow clients to create records of the respective model.
```
nikolasburk: maybe emphasize the "one"-aspect of this operation with: "Allow clients to create single records of the respective model."

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+Relation fields may be connected with an existing record or a sub-create may be inlined. If the relation is a `List` then multiple connections or sub-creates are permitted.
```
nikolasburk: is `List` an actual programmatic concept in this context? otherwise I suggest lowercase spelling and no code formatting (if it is, is there maybe a link to docs of what that `List` type is`?

nikolasburk: > Relation fields may be connected with an existing record or a sub-create may be inlined.

I really like this sentence! I wonder if we somehow should throw in the term "nested mutation". This is the term we've used to describe these features in the Graphcool/Prisma GraphQL APIs and I think it's pretty intuitive. One idea would be to add it in parens after the sentence: "Relation fields may be connected with an existing record or a sub-create may be inlined (_nested mutations_)."

nikolasburk: or adding an extra sentence to the end of the paragraph: "This is also referred to as _nested mutations_."

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+Each instance of this operation contributions to GraphQL Schema:
```
nikolasburk: lowercase "schema"

nikolasburk: "contributions" -> should that be "contributes"?

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+- `×1` `InputObject` `<ModelName>CreateInput`
```
nikolasburk: the explanation in all the bullets doesn't feel intuitive to me. I think I understand but I imagine this to be unclear/confusing for newcomers... also the 3 times incline code formatted words without context probably aren't the best way to describe this. How about trying to put this information into a table instead and adding some more context/description. (I think much of the confusion can probably be resolved by different formatting and adding a bit of context.)

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+  This is a container for the sub-create and connect (below). Cannot be amortized becuase sub-create cannot.
+
+- `×N` `InputObject` per relational field create.
+
+  Cannot be amoritized because of specialized shape excluding field where supplying relation of type of parent `Object` being created would normally be. Instead this relation being created will be intuitively connected to the `Object` being directly created.
+
+- `×N (A)` `InputObject` `<ModelName>WhereUniqueInput` per relational field connect.
+
+Example:
```
nikolasburk: notes for the example:

- I think it would be good to prepend this with a Prisma schema the example is based on
- add at one sentence or a short description before each code block to explain what this is/where it comes from
- maybe spell **Example** in bold here or make it a `#####` heading?

nikolasburk: Ah just realized there's a Prisma schema at the end, I'd lead with it since it sets the stage for everything that follows?

README.md
nikolasburk: also, I prefer remaining explicit and referring to it as "Prisma model"

README.md
```diff
@@ -5,11 +5,1119 @@
...
+  id     Int    @id
+  title  String @unique
+  body   String
+  author User
+}
+```
+
+#### One-Read Operation
+
+Allow clients to find a specific record of the respective model. They may search by any field in the model marked with `@unique` attribute.
```
nikolasburk: again, I prefer remaining explicit and referring to it as "Prisma model"

README.md
```diff
@@ -5,11 +5,1119 @@
...
+
+The ability for relation fields to be filtered sorted etc. depends on if such features have been enabled for those `Object`s [via `t.model`](/TODO).
```
nikolasburk: > The ability for relation fields to be filtered sorted etc.

I suggest to spell out pagination as third feature: "The ability for relation fields to be filtered, sorted and/or paginated depends on ..."

nikolasburk (changes requested):
I think this looks really great @jasonkuhrt!! 💯I've made a number of comments to share my own opinions, mostly regarding formatting, spelling, terminology etc. Also happy to jump on a call to discuss some of the feedback if necessary 🙂

README.md
jasonkuhrt: I have to make doctoc + vsc happy. doctoc ignores titles above the toc, and vsc permits a comment to ignore titles 👍 

Are you ok with an H3? I feel H1 competes with the title and H2 having the horizontal bar adds too move "bar noise" at the top. Also a bit of context: I was thinking the opening text should be pretty minimal and not actually constitute a section in its own right. Just a bit of feeler text. Having it just flow casually into the contents rather than it being these big bold distinct sections.

image

image

image

README.md
jasonkuhrt: > add at one sentence or a short description before each code block to explain what this is/where it comes from

The reference examples follow a repeated pattern that aims to be lean. The goal is illustrating the information across 4 layers related to that operation. It is abstract with regards to how an app is formed or where the files live or what client is being used. For example I don't bother with `export const Post =` because it just adds noise. It doesn't help make the point I think. Maybe a better heading here (than Example) would be Illustration. Might be neat if each one had a link to a "try now" codesandbox. Future idea maybe.

> I'd lead with it since it sets the stage for everything that follows?

It's in order of the stack layers, but we could reverse. Right now its in a literate programming style order. Beginning with the result, working backward to the implementation. This approach allows one to ask "How is that working?, How is that working?, ..." vs "Why this/what happens next?, Why this/what happens next? ...". 

Let's revisit this once the content is done and see how it feels overall. Not sure yet.

README.md
jasonkuhrt: > I'm a bit confused by the inline code formatting for Object? are you referring to objectType here? I think for inline code formatting, we should try to apply it only to concepts that are literally code pieces, so if Object stands for objectType here, maybe indeed use objectType instead?

It is the formal (spec) GraphQL Type name. Along with `Enum` `InputObject` `Union` and so on. A verbose way of saying it might be "GraphQL `Object`". Situations like this might be reason for us to create a small legend/intro/glossary.

Here is a before/after comparison:

Before:

image

After:

image

Will think about this more!

> Question: Are we sure we want to uppercase concepts like field projectors? To me these are regular features (not names) so I'd rather lowercase them. I've been in that situation a number of times when documenting Prisma features, and I feel like defaulting to lowercasing generally feels more natural. Otherwise it's hard to draw a boundary when something gets lowercased and when it gets uppercased.

Yeah not sure. Will think about this more!

> Second note: I'd write this in italics instead of "quotes". 

👍

README.md
jasonkuhrt: > Again, I suggest to be a bit more explicit and call out that model is a Prisma concept: "t.model will have Field Projectors for the Prisma model whose name matches that of the Object"

👍

README.md
jasonkuhrt: (mentioned somewhere else my feelings about that)

README.md
jasonkuhrt: (somewhere you mentioned realizing that the PSL is mentioned, just last not first. And we're discussing the order now in that thread)

README.md
jasonkuhrt: (having a discussion about this in another thread)

README.md
jasonkuhrt: > maybe again be explicit that Query and Mutation are Nexus concepts, e.g. with: It is only available inside the Query and Mutation types that are defined with the GraphQL Nexus API.

I've tried to work with the GraphQL terminology. In Nexus there are multiple ways that one could get to a definition block for `Query`:

```
queryType({ ... })
```
```
extendType({ name: "Query",  ... })
```
```
objectType({ name: "Query",  ... })
```

While a truly exhaustive reference would probably state every last detail like this I'm trying to find a tradeoff both for ourselves and the user (who might be able to get the point without facing so much text).

I do think we should mention this, but just once, someone centralized, will think about it!

README.md
jasonkuhrt: Something I picked up from my uni days, first link I found in Google

> A simple rule for using numbers in writing is that small numbers ranging from one to ten (or one to nine, depending on the style guide) should generally be spelled out. Larger numbers (i.e., above ten) are written as numerals.

–– https://www.scribendi.com/advice/when_to_spell_out_numbers_in_writing.en.html

I feel like the adjective "exactly" doesn't add much value, I'll just try writing the numbers out as digits.

README.md
jasonkuhrt: Also something I'm trying to be careful with.  Sometimes we talk about GraphQL Objects. Sometimes we talk about Prisma Models. It's difficult to talk about the layers clearly. We even have [problems in the API related to that](https://github.com/prisma-labs/nexus-prisma/issues/460).

Should this sentence instead say:

> Allow clients to create one record of the respective GraphQL Object.

As we get into [distributed data](https://github.com/prisma-labs/nexus-prisma/issues?q=is%3Aopen+is%3Aissue+label%3Ascope%2Fdistributed-data), clients may work with Objects that don't even map to a single Prisma model. Maybe One GraphQL Object is 3 Prisma models or 2 Prisma Models and one Stripe API GraphQL Object... 

That said, the crud operations API of nexus-prisma right now only support on 1:1 Prisma model  <> Graphql Object. So maybe that sentence is alright.

README.md
nikolasburk: I actually also prefer spelling out numbers instead of using digits (even for numbers up to hundred), so you don't have to switch to digits for me. Thinking more about this, maybe the "exactly" isn't too important,... IMO it would add _a bit_ of value since it points out that this number is very intentional at this point, it's hard to describe for me but otherwise reading this sentence the ten feels a bit random/lost/out-of-place 🙈 no strong opinion though, so feel free to leave it as is!

README.md
jasonkuhrt: > is List an actual programmatic concept in this context? otherwise I suggest lowercase spelling and no code formatting (if it is, is there maybe a link to docs of what that List type is`?

Similar to what I mentioned else where, a GraphQL Type.

> (nested mutations)

👍

README.md
jasonkuhrt: Will think about how to simply, try table etc.

jasonkuhrt:
@nikolasburk thanks so much for all your feedback, I've incorporated most of it already.

I have done several major changes architecturally and expect a few more coming.

jasonkuhrt:
For posterity, and if someone wants to discuss alternatives to the GQL contributions notation, here are various notations I explored before settling on the current one.

Screen Shot 2019-10-11 at 11 12 46 PM
Screen Shot 2019-10-11 at 2 52 15 PM
Screen Shot 2019-10-11 at 2 39 36 PM
Screen Shot 2019-10-11 at 2 27 32 PM
Screen Shot 2019-10-11 at 2 27 27 PM
Screen Shot 2019-10-11 at 2 26 40 PM
Screen Shot 2019-10-11 at 2 26 29 PM
image

PR #487, opened 2019-10-15
fix: publish scalars from input types

### LINKED ISSUE #486: filtering option should auto create GQL custom scalars
#### Example

```prisma
model User {
  id String   @id @unique @default(cuid())
  d DateTime
}
```
```ts
objectType({
  name: 'User',
  definition(t) {
    t.model.id()
  },
})
  queryType({
    definition(t) {
      t.crud.users({ filtering: true })
    },
  })
```
```gql
### This file was autogenerated by GraphQL Nexus
### Do not make changes to this file directly


input DateTimeFilter {
  equals: NEXUS__UNKNOWN__TYPE
  gt: NEXUS__UNKNOWN__TYPE
  gte: NEXUS__UNKNOWN__TYPE
  in: [NEXUS__UNKNOWN__TYPE!]
  lt: NEXUS__UNKNOWN__TYPE
  lte: NEXUS__UNKNOWN__TYPE
  not: NEXUS__UNKNOWN__TYPE
  notIn: [NEXUS__UNKNOWN__TYPE!]
}

type Mutation

"""

    This scalar should never make it into production. It is used as a placeholder for situations
    where GraphQL Nexus encounters a missing type. We don't want to error immedately, otherwise
    the TypeScript definitions will not be updated.
  
"""
scalar NEXUS__UNKNOWN__TYPE

type Query {
  users(after: String, before: String, first: Int, last: Int, skip: Int, where: UserWhereInput): [User!]!
}

input StringFilter {
  contains: String
  endsWith: String
  equals: String
  gt: String
  gte: String
  in: [String!]
  lt: String
  lte: String
  not: String
  notIn: [String!]
  startsWith: String
}

type User {
  id: ID!
}

input UserWhereInput {
  AND: [UserWhereInput!]
  dd: DateTimeFilter
  id: StringFilter
  NOT: [UserWhereInput!]
  OR: [UserWhereInput!]
}

```
```
Error: 
- Missing type DateTime, did you forget to import a type to the root query or mean DateTimeFilter?
```

#### Use-Case

User might not want to project a Prisma scalar field onto a GQL Object _but_ want to be able to filter by that Prisma scalar field.

#### Work-around

1. Project `t.model.d()`
2. Define custom scalar in GQL schema

#### Notes

- Auto-creating custom GQL scalars is behaviour we have today for when projecting Prisma scalar fields.
- Example shows DateTime but applies to any Prisma scalar that is a custom Scalar in GQL (e.g. UUID)

Weakky:
Fixes #486

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ example/generated
 .DS_Store
 yarn-error.log
 failed-*.md
+.idea
--- a/src/publisher.ts
+++ b/src/publisher.ts
@@ -2,7 +2,7 @@ import * as Nexus from 'nexus'
 import { CustomInputArg } from './builder'
 import * as DMMF from './dmmf'
 import { scalarsNameValues } from './graphql'
-import { dmmfFieldToNexusFieldConfig, partition } from './utils'
+import { dmmfFieldToNexusFieldConfig } from './utils'
 
 export class Publisher {
   constructor(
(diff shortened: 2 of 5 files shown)


jasonkuhrt (approved):
Nice, solution actually made the code simpler too :)

PR #521, opened 2019-11-05
fix: updates necessary for nexus 0.12.0

tgriesser:
Moves the changes & tests for `NEXUS_SHOULD_GENERATE_ARTIFACTS` into this plugin if you still want them. Otherwise just double checking that things still work. Saw some issue in the tests loccally around the argument gen for `t.crud.users` being incorrect, not sure how that works or if it's related to the changes in 0.12:

Screen Shot 2019-11-05 at 10 28 23 AM


jasonkuhrt:
@tgriesser Many thanks! I'm not sure we're going to put this logic into this project or not, but either way its a great reference for us now. 🙌

jasonkuhrt (approved):

PR #532, opened 2019-11-17
Monorepo support by resolving generated types

kripod:
When multiple 'node_modules' directories are in use, using `require.resolve` looks up the `@generated/photon` package recursively.


kripod:
Superseded by #534.

PR #533, opened 2019-11-17
fix(typegen): normalize generated path on windows

kripod:
Previously, the generator would output statements like:

```ts
import * as photon from 'D:\Development\Projects\app\server\node_modules\@generated\photon\index.js';
```

Which is invalid, as the `\` characted should be escaped to `\\`. Luckily, Windows also accepts `/` as a separator, providing a universal solution compatible with all the platforms:

```ts
import * as photon from 'D:/Development/Projects/app/server/node_modules/@generated/photon/index.js';
```


kripod:
Superseded by #536.

PR #534, opened 2019-11-18
feat: add monorepo support by resolving a path

kripod:
When multiple 'node_modules' directories are in use, using `require.resolve` looks up the `@generated/photon` package recursively.


jasonkuhrt (changes requested):
Hey @kripod could you rebase with master which now is designed around the `@prisma/photon` facade.

PR #535, opened 2019-11-19
feat: prevent errors if a wrong field is projected

### LINKED ISSUE #472: Use proxies to avoid not defined error blocking compile
Analogy: Unused variable as error prevents build. At dev time this is annoying because an unused variable may just be a temporary state while working towards some idea, refactor, etc. It is much better at dev time to have this feedback at the level of just a warning, that does not block the compiler from building your code.

The idea is to implement the `t.model` operation publishers and `t.crud` field projectors as proxies so that when there is a sync issue or typo TS compile can still succeed. With proxies we can:

- Provide custom error messages when the field doesn't exist (`please sync` etc.)
- never get in the way of TS compiling, a golden rule

We may want a prod mode where proxies aren't used.

Weakky:
- Only when NODE_ENV !== 'production'
- Add a hook for pumpkins. If such hook is present, it lets the hook log the error

Fixes #472

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -13,6 +13,7 @@ import {
 import { Publisher } from './publisher'
 import * as Typegen from './typegen'
 import { assertPhotonInContext } from './utils'
+import { proxify, OnUnknownFieldName } from './proxifier'
 
 interface FieldPublisherConfig {
   alias?: string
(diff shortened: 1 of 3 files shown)


jasonkuhrt (approved):
<3 ⛵️

PR #536, opened 2019-11-19
fix(typegen): support serverless and windows

kripod:
Previously, the type generator would output statements like:

```ts
import * as photon from 'C:\Development\Projects\exigo\server\src\functions\graphql\.generated\photon';
```

Which is invalid, as the `\` characted should be escaped to `\\`. Luckily, Windows also accepts `/` as a separator, providing a universal solution compatible with all the platforms:

```ts
import * as photon from 'C:/Development/Projects/exigo/server/src/functions/graphql/.generated/photon';
```

In addition to this, Serverless platforms (e.g. Netlify Functions) may not support absolute paths, so the module path given above should be resolved relative to the containing directory of the generated types, e.g.:

```ts
import * as photon from './.generated/photon/index.js';
```

Given the schema configuration below:

```ts
import { makeSchema } from 'nexus';
import { nexusPrismaPlugin } from 'nexus-prisma';
import * as path from 'path';

export default makeSchema({
  types: [/* ... */],
  plugins: [
    nexusPrismaPlugin({
      inputs: {
        photon: path.join(__dirname, '../.generated/photon'),
      },
      outputs: {
        typegen: path.join(__dirname, '../nexus-prisma.generated.d.ts'),
      },
    }),
  ],
  outputs: {
    typegen: path.join(__dirname, '../nexus.generated.d.ts'),
  },
});
```

### DIFF
--- a/src/typegen.ts
+++ b/src/typegen.ts
@@ -31,7 +31,6 @@ export function doGenerate(
   options: Options,
 ): void | Promise<void> {
   const dmmf = DMMF.get(options.photonPath)
-  console.log({ options })
   const tsDeclaration = render(
     dmmf,
     isModule(options.photonPath)
--- a/src/typegen.ts
+++ b/src/typegen.ts
@@ -1,3 +1,4 @@
+import * as path from 'path'
 import * as DMMF from './dmmf'
 import { getCrudMappedFields } from './mapping'
 import { defaultFieldNamingStrategy } from './naming-strategies'
@@ -23,7 +24,14 @@ export function doGenerate(
   options: Options,
 ): void | Promise<void> {
   const dmmf = DMMF.get(options.photonPath)
-  const tsDeclaration = render(dmmf, options.photonPath)
+  const tsDeclaration = render(
+    dmmf,
+    // Support Serverless platforms by resolving module paths relatively
+    './' +
+      path
+        .relative(path.dirname(options.typegenPath), options.photonPath)
+        .replace(/\\/g, '/'), // Use '/' instead of '\' on Windows
+  )
   if (sync) {
     hardWriteFileSync(options.typegenPath, tsDeclaration)
   } else {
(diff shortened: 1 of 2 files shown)


src/typegen.ts
```diff
@@ -16,14 +17,31 @@ export function generate(options: Options): Promise<void> {
...
+
 export function doGenerate(sync: true, options: Options): void
 export function doGenerate(sync: false, options: Options): Promise<void>
 export function doGenerate(
   sync: boolean,
   options: Options,
 ): void | Promise<void> {
   const dmmf = DMMF.get(options.photonPath)
-  const tsDeclaration = render(dmmf, options.photonPath)
+  console.log({ options })
```
jasonkuhrt:
```suggestion
```

src/typegen.ts
```diff
@@ -16,14 +17,31 @@ export function generate(options: Options): Promise<void> {
   return doGenerate(false, options)
 }
 
+function isModule(name: string) {
```
jasonkuhrt:
```suggestion
function isPackage(name: string) {
```

This seems to leak in the face of e.g. `from 'fs-jetpack/foo`? I think you can go further in the guard by checking for leading `./`.

src/typegen.ts
```diff
@@ -16,14 +17,31 @@ export function generate(options: Options): Promise<void> {
...
+  console.log({ options })
+  const tsDeclaration = render(
+    dmmf,
+    isModule(options.photonPath)
+      ? options.photonPath
+      : // Support Serverless platforms by resolving module paths relatively
+        './' +
+          path
+            .relative(path.dirname(options.typegenPath), options.photonPath)
+            .replace(/\\/g, '/'), // Use '/' instead of '\' on Windows
```
jasonkuhrt, line 42: A windows util func might be warranted here, like you did for `quoteOnWindows`. Would be re-used in the test suite too.

jasonkuhrt (changes requested):

src/typegen.ts
kripod: Thank you for the suggestion! Unfortunately, I don’t have access to a computer right now, but feel free to modify the PR.

PR #539, opened 2019-11-21
feat: add support for latest prisma 2

jasonkuhrt:


package.json
```diff
@@ -20,14 +20,17 @@
     "test": "jest"
   },
   "dependencies": {
-    "@prisma/photon": "0.2.152",
-    "@prisma/sdk": "0.0.23",
+    "@prisma/photon": "^2.0.0-alpha.344",
+    "@prisma/sdk": "^0.0.55",
     "camelcase": "^5.3.1",
     "fs-extra": "^8.1.0",
+    "js-levenshtein": "^1.1.6",
```
jasonkuhrt: Hack

This is a transient dep but not being pulled in during install correctly. IIUC Photon needs to fix its manner of depending on `js-levenshtein`.

package.json
```diff
@@ -20,14 +20,17 @@
...
     "nexus": "^0.12.0-rc.4",
-    "pluralize": "^8.0.0"
+    "pluralize": "^8.0.0",
+    "prisma2": "^2.0.0-alpha.344"
   },
   "devDependencies": {
+    "@types/debug": "^4.1.5",
```
jasonkuhrt: Hack

This is a transient dep but not being pulled in during install correctly. IIUC Photon needs to fix its manner of depending on `@types/debug`.

divyenduz (approved):
PR looks good, since the release is happening today, maybe we can already fix these hacks in the source (Photon)

PR #540, opened 2019-11-23
feat(crud): computed inputs

ssalbdivad:
nexus-prisma and the prisma ecosystem can remove a huge amount of the boilerplate that is typically required to build a scalable backend. However, a few weeks ago I was working on some user-specific types and found myself writing a lot of redundant code.

For example, assume each User has a one-to-many relationship with Post and Photo. To create resolvers that can create a Post or Photo for the User that is currently authorized, you're stuck either:

- Including User in your API's InputType, which it shouldn't be since it can be determined from context
- Writing custom resolvers and InputTypes for createOnePost and createOnePhoto, which is doable but adds a lot of boilerplate and feels like it should be unnecessary for a simple case like this

I asked about this in the prisma Slack channel a few weeks ago, and I was told there wasn't a good solution for this currently but that I should log an issue. I ended up implementing a solution that adds a new "contextArgs" option for nexus-prisma users in a similar situation.

Here's an example of how it's used, which I've also included as a new unit test:

```
model User {
    id  Int @id
    name String
    browser String
}
...
const Mutation = Nexus.objectType({
    name: 'Mutation',
    definition(t: any) {
      t.crud.createOneUser({
        contextArgs: { browser: (ctx: any) => ctx.browser },
      })
    },
})
```

Imagine in this case, we have an app in which Users have a "browser" field that stores the browser with which their account was created. However, we don't want to have to include that in the GraphQL API because we can get it from session info and put it in context. By passing a contextArgs object containing browser, we ensure that:

- The UserCreateInput type in the GraphQL schema omits the "browser" field
- Each time createOneUser is executed, whatever args are passed (in this case just "name") are merged with args derived from context, i.e. the keys of the object passed to contextArgs. The values of the object are functions that accept context as their only parameter and return the value of the arg that should be passed to the Photon method. In the example, the resolution of createOneUser looks for a "browser" key in context and then passes it to Photon, however there is no requirement that the data fetched from context has the same name as contextArg it ends up in. For example, we could instead have a "session" key in context from which we extract browser, but it would still map to a "browser" key in the contextArg object because that's where we want it to end up when we're passing args to Photon.

In addition to creating a new unit test, I was able to test this out with the blog example included in the nexus-prisma repo and although I'm new to the project, it seems to be working well. While learning about how nexus-prisma is organized, I also took the opportunity to refactor some redundant code in the builder. Any feedback would be greatly appreciated, and I hope this is helpful to other users! Thanks so much for all the work you've done; the developer experience is second to none and I can't wait to see what comes next 😃


ssalbdivad:
@schickling Is anyone able to take a look at this in the next couple days? Would love to be able to consume the change from the package directly so I can continue to adopt new releases!

jasonkuhrt:
Hey @ssalbdivad thanks so much for coming with this amount of work and thought. We're going to give this PR a proper look and feedback next sprint, which for us starts tomorrow.

My goal with the Labs team is to respond to this kind of thing within 24h, which didn't happen here. That's on me. Thanks for your patience 🙏

ssalbdivad:
@jasonkuhrt No problem on the delayed response. The extra time actually let me consume some of the changes in our app and see how they played out in the real world. I found that it was a huge improvement to be able to omit fields like user that can be inferred from context from a mutation's input type, but I was still frustrated that I had to write custom resolvers if a nested type required the same input.

Since the original PR I built a second feature that allows users to specify global context args to solve that problem.

I've included documentation in my PR which I'll add in a separate comment for reference. I've also written several new unit tests in addition to testing the change from my own app. Given this change is totally opt in and wouldn't impact existing users, it'd be hugely appreciated if someone on your team could take a look this sprint. Admittedly it does touch a lot of code, particularly in builder.ts, so I'm worried about merge conflicts if a lot of other work is done first. Let me know if there is anything I can do to help the process along, and thanks again for your team's great work!

ssalbdivad:
### `contextArgs` (resolver-level)

```
Record<string, (ctx: any) => any> | undefined
```

**Applies To**

[`t.crud.<mutations>`](#t.crud)

**About**

Allow clients to omit fields from one mutation's corresponding input type and infer the value of those fields from context when determining what to pass to Photon.

- `undefined` (default) Like `{}`; inputs are unchanged
- `ContextArgs` (`Record<string, (ctx: any) => any>`) Keys in the ContextArgs object will be omitted from the mutation's corresponding input type. When resolving the mutation at runtime, each omitted key will be passed to Photon based on the return value of that key's corresponding function in the ContextArgs object when passed the current request's context.

**GraphQL Schema Contributions**

The mutation's input type fields with a name that is in the ContextArgs object are omitted from the GraphQL schema. This modifies one existing input type but does not add new types or remove existing types.

**Example**

```prisma
model User {
  id  Int @id
  name String
  createdWithBrowser String
}
```

```ts
queryType({
  definition(t: any) {
    t.crud.user()
  },
})
mutationType({
  definition(t: any) {
    t.crud.createOneUser({
      contextArgs: {
        createdWithBrowser: (ctx: any) => ctx.browser,
      },
    })
  },
})
objectType({
  name: 'User',
  definition: (t: any) => {
    t.model.id()
    t.model.name()
    t.model.createdWithBrowser()
  },
})
```

```gql
type Mutation {
  createOneUser(data: UserCreateInput!): User!
}

type Query {
  user(where: UserWhereUniqueInput!): User
}

type User {
  id: Int!
  name: String!
  createdWithBrowser: String!
}

input UserCreateInput {
  name: String!
}

input UserWhereUniqueInput {
  id: Int
}
```

```gql
mutation createOneUser {
  createOneUser({data: {name: "Benedict"}}) {
    id
    name
    createdWithBrowser
  }
}
```

<br>

### `contextArgs` (global)

```
Record<string, (ctx: any) => any> | undefined
```

**Applies To**

[`nexusPrismaPlugin()`](#Example)

**About**

Allow clients to omit fields with a given name across all of their GraphQL schema's inputs and infer the value of those fields from context when determining what to pass to Photon

- `undefined` (default) Like `{}`; inputs are unchanged
- `ContextArgs` (`Record<string, (ctx: any) => any>`)
  Keys in the ContextArgs object will be omitted from all input types. When resolving the mutation at runtime, each omitted key will be passed to Photon based on the return value of that key's corresponding function in the ContextArgs object when passed the current request's context recursively.

**GraphQL Schema Contributions**

All input type fields with a name that is in the ContextArgs object are omitted from the GraphQL schema. This modifies existing input types but does not add new types or remove existing types.

**Example**

```prisma
model User {
  id  Int @id
  name String
  nested Nested[]
  createdWithBrowser String
}

model Nested {
  id Int @id
  name String
  createdWithBrowser String
}
```

```ts
queryType({
  definition(t: any) {
    t.crud.user()
  },
})
mutationType({
  definition(t: any) {
    t.crud.createOneUser()
    t.crud.createOneNested()
  },
})
objectType({
  name: 'User',
  definition: (t: any) => {
    t.model.id()
    t.model.name()
    t.model.nested()
    t.model.createdWithBrowser()
  },
})
objectType({
  name: 'Nested',
  definition: (t: any) => {
    t.model.id()
    t.model.createdWithBrowser()
    t.model.name()
  },
})
```

```ts
nexusPrismaPlugin({
  /*
  Remove fields named "createdWithBrowser" from all mutation input types. When resolving
  a request whose data contains any of these types, the value is inferred
  from context and passed to Photon, even if it's nested. Assumes a Photon
  context containing a "browser" key that maps to a string representing
  the browser the request was made from.
  */
  contextArgs: {
    createdWithBrowser: ctx => ctx.browser,
  },
})
```

```gql
type Mutation {
  createOneUser(data: UserCreateInput!): User!
  createOneNested(data: NestedCreateInput!): Nested!
}

type Nested {
  id: Int!
  createdWithBrowser: String!
  name: String!
}

input NestedCreateInput {
  name: String!
  user: UserCreateOneWithoutUserInput
}

input NestedCreateManyWithoutNestedInput {
  create: [NestedCreateWithoutUserInput!]
  connect: [NestedWhereUniqueInput!]
}

input NestedCreateWithoutUserInput {
  name: String!
}

input NestedWhereUniqueInput {
  id: Int
}

type Query {
  user(where: UserWhereUniqueInput!): User
}

type User {
  id: Int!
  name: String!
  nested(skip: Int, after: Int, before: Int, first: Int, last: Int): [Nested!]!
  createdWithBrowser: String!
}

input UserCreateInput {
  name: String!
  nested: NestedCreateManyWithoutNestedInput
}

input UserCreateOneWithoutUserInput {
  create: UserCreateWithoutNestedInput
  connect: UserWhereUniqueInput
}

input UserCreateWithoutNestedInput {
  name: String!
}

input UserWhereUniqueInput {
  id: Int
}
```

```gql
mutation createOneUser {
  createOneUser({data: {name: "Benedict", nested: {name: "Moony"}}) {
    id
    name
    createdWithBrowser
    nested {
      id
      name
      createdWithBrowser
    }
  }
}

mutation createOneNested {
  createOneNested({data: {name: "Moony", user: {connect: {where: {id: 1}}}}}) {
    id
    name
    createdWithBrowser
  }
}
```

If `{user: {connect: {where: {id: 1}}}}` looks familiar, global contextArgs can also be used to determine the user making a request and automatically populate mutations affecting a single user accordingly. For example, assuming Photon's context includes a "userId" key, adding a user key to global contextArgs can simplify the "createdOneNested" mutation from the previous example:

```ts
nexusPrismaPlugin({
  ...other config...
  contextArgs: {
    createdWithBrowser: ctx => ctx.browser,
    user: ctx => ({ connect: { where: { id: ctx.userId } } }),
  },
})
```

```gql
mutation createOneNested {
  createOneNested({data: {name: "Moony"}}) {
    id
    name
    createdWithBrowser
  }
}
```

jasonkuhrt:
@ssalbdivad have now had time to skim the value proposition, use-cases, docs. Thanks for making it so easy to review this PR from a product perspective. This is great work 🙏.

I like this feature. Its declarative and feels consistent with the vision and API of nexus-prisma. CRUD customization is something that @Weakky and I went deep on during a brainstorm week this past October in Berlin (we're both remote). We discovered a lot of API/product issues/opportunities during this time. While our efforts since then were diverted to some other goals (workflow related) we know CRUD is key and know we've only touched the tip of the iceberg so far. Anyways point is that your API ideas here align what some of the observations we also made. For example the onion layer of local-to-global declarative config––resolver, global. The graph nature of the CRUD input types is a challenge to abstract well. But I feel this feature provides a lot of value, is intuitive, at a small cost to the cognitive load of using the CRUD DSL. It also helps that it is additive, making it that much easier to adopt this feature.

I am going to discuss more with @Weakky soon. Probably Friday as he is busy preparing for the Berlin GQL meetup today/tomorrow.

🙌

jasonkuhrt:
Hey @ssalbdivad, said discussion is happening tomorrow.

ssalbdivad:
@jasonkuhrt Great! Let me know if I can provide any additional information about the PR. I know there are a lot of changes in builder.ts in particular but the vast majority of the code is functionally the same with just a bit of refactoring to keep my changes DRY.

jasonkuhrt:
Hey @ssalbdivad here's the outcome of our talk today;

We considered these alternatives:

1. Declarative CRUD visibility + procedural implementation

    ```ts
    objectType({
      name: "Mutation",
      definition(t) {
        t.createOneUser({
          // args: Omit<args, 'createdWithBrowser'>
          customResolver(root, args, ctx, info, oldResolver) {
            const oldResolverArgs = {
              ...args,
              createdWithBrowser: ctx.browser
            }
            return oldResolver(root, oldResolverArgs, ctx, info);
          }
        });
      }
    });

    objectType({
      name: "User",
      definition(t) {
        t.model.id();
        t.model.name();
        t.model.createdWithBrowser({ hide: ["create"] });
      }
    });
    ```

2. Declarative computed inputs

    ```ts
    objectType({
      name: "Mutation",
      definition(t) {
        t.createOneUser({
          computedInputs(root, args, ctx) {
            return {
              createdWithBrowser: ctx.browser,
            };
          }
        });
      }
    });

    objectType({
      name: "User",
      definition(t) {
        t.model.id();
        t.model.name();
        t.model.createdWithBrowser();
      }
    });
    ```

3. Declarative computed inputs + contextArgs static helper

    ```ts
    objectType({
      name: "Mutation",
      definition(t) {
        t.createOneUser({
          computedInputs: contextArgs({
            createdWithBrowser: ctx => ctx.browser
          })
        });
      }
    });

    objectType({
      name: "User",
      definition(t) {
        t.model.id();
        t.model.name();
        t.model.createdWithBrowser();
      }
    });
    ```

Each of the above could be seen as increasingly high level. Successive sugar over the previous iteration. We saw your proposal like a fourth layer of sugar, where the "computed inputs" are "hardcoded" to come from context.

Considering we do not yet have any of these levels established yet, we're hesitant to start with the fourth. Not that we're against working backwards form high-level in general––but in this case its what we felt.

We think starting with layer 1 or 2 might be better.

Layer 1 is part of a larger CRUD vision, clearly, which introduces the concept of, and centralizes, model field visibility in CRUD at the point of model field projection. This is a general direction we're probably going to be exploring sooner than later (months scale).

Layer 2 would somehow be like implying layer 1. But rather than two steps (declare + implement) the implementation would double as declaration of intent. This has the downside of offering less contract-driven control of the DX, and feedback our system can give the user. But its more direct in some ways for the user.

Would be interested in any feedback you have about our thoughts! Hope its not a joykill. The good news is we're definetely seeing the same end goal(s), its just a question of how we get there.

CC @Weakky jump in if I missed or misrepresented anything 🙌

ssalbdivad:
@jasonkuhrt Thanks for taking the time to review this in such detail, the feedback is very helpful! I think having a more configurable "computedInputs" type option makes a lot of sense. However, having consumed the contextArgs change in my app, I also feel the improvement in DX would be really painful to give up in the meantime. Redo is an app that creates automated tests for your site by watching you interact with it. Here's a simplified version of our schema.prisma:

```
model Tag {
  id   Int    @id
  user User
  name String
  @@unique([name, user])
}

model Selector {
  id  Int    @id
  css String
}

model Step {
  id       Int      @id
  action   String
  selector Selector
  value    String
}

model Test {
  id    Int    @id
  user  User
  name  String
  steps Step[]
  tags  Tag[]
  @@unique([name, user])
}

model User {
  id        Int        @id
  email     String     @unique
  password  String
  first     String
  last      String
  steps     Step[]
  selectors Selector[]
  tags      Tag[]
  tests     Test[]
}
```

Using a globalContext arg to infer the value of "user" inputs from context like this:
```
nexusPrismaPlugin({
    contextArgs: {
        user: ({ userId }) => ({
            connect: {
                id: userId
            }
        })
    }
})
```
I can just add ```t.crud.createOneTest()``` to my mutationType definition and I get a working mutation on my GraphQL server that I can use like this:

```
mutation createTest {
    createOneTest(
        data: {
            name: "Test 1"
            tags: { create: [{ name: "critical" }] }
            steps: {
                create: {
                    action: "click"
                    value: "something"
                    selector: { create: { css: "#someId" } }
                }
            }
        }
    ) {
        id
    }
}
```

Even if we had custom revolvers, given that Test, Tag, Step and Selector are all correlated with a User, a basic create mutation for any of these will require a custom input type and custom resolver, which takes that single ```t.crud.createOneTest()``` line of code and turns it into many dozens. It was in initially writing all that extra boilerplate that I felt there must be a better way and started working on this change, but I think for a lot of new users it would be a huge roadblock.

I'm very much looking forward to the custom resolver work you outlined. In the meantime though, I don't want to be maintaining hundreds of lines of code that can be summarized as "get fields called 'user' from context instead of requiring them in the API." It also seems like having lots of data tied to user accounts would be common across many if not most apps. While contextArgs could eventually be integrated with a customization strategy like the one you described, the benefit of making this change available to users during the intervening months seems to far outweigh the potential confusion that could come along with potentially deprecating an old version of contextArgs in favor of one that fits into a larger cohesive strategy. Let me know your thoughts, and thanks again for the feedback! I'd love to work with you all on the next stages when the time is right :bowtie:

ssalbdivad:
@jasonkuhrt Any thoughts on this? I'm still consuming my forked version of the package and would really like to be able to consume this alongside the new changes you guys are working on.

jasonkuhrt:
Hey @ssalbdivad, thanks for the continued feedback. We're racing toward a deadline  for next week so a bit hectic. Totally hear you on not wanting the burden of fork maintenance etc. Will think  a bit more about next steps. Even considering if you jump on a call with @Weakky and I. We'll see, still need to hit that deadline :D 

Also any chance you'd be open to pursing an implementation of some of the specs I began to sketch out above? Maybe level 2 for example. Not sure, just curious on your bandwidth/interest levels here.

Talk soon!

ssalbdivad:
@jasonkuhrt Thanks for getting back so soon! Admittedly I spent more time on the initial change than I should have given I'm in the early stages of building my own startup, but most of that was getting a mental model of the code base. At first glance, the level 2 changes *might* be a fairly straightforward adaption of what I've done so far. I'll take another look tonight and if it seems like something I can bang out in a couple hours I'll give it a shot, otherwise I still want to make sure we don't end up with the perfect as the enemy of the good :bowtie:

ssalbdivad:
@jasonkuhrt Latest commits include a "computedInputs" options similar to what you described. The only difference is that rather than a single function returning an object, I needed to use an object that maps to individual functions for each key so that we can determine the fields to remove during typegen. While working on the change, I was finding it difficult to keep track of all of the similar sounding aliases in the "dmmf" directory, so I also moved/renamed a few things which I hope makes that part of the code more intuitive. Updated the unit tests, tested end-to-end with Redo's server, and updated the documentation, which I'll include in my next comment. Let me know what you think! :bowtie:

ssalbdivad:
### `computedInputs` (local)

```
Record<string, ({root, args, ctx}: MutationResolverParams) => any> | undefined
```

Note: This is an abbreviated version of the ComputedInputs type. The most important thing to undertand each of the object's values will be a function that takes an object with "root", "args" and "ctx" keys that represent the runtime values of the corresponding parameters that are passed to your resolver. For the full type, see [ComputedInputs Type Details](#computedinputs-type-details).

**Applies To**

[`t.crud.<mutations>`](#t.crud)

**About**

Allow clients to omit fields from one mutation's corresponding input type and infer the value of those fields from the resolver's params (root, args, and context) at runtime when determining what to pass to Photon.

- `undefined` (default) Like `{}`; inputs are unchanged
- `ComputedInputs` (`Record<string, ({root, args, ctx}: MutationResolverParams) => any> | undefined`) [(full type here)](#computedinputs-type-details).

  Keys in the ComputedInputs object will be omitted from the mutation's corresponding input type. When resolving the mutation at runtime, each omitted key will be passed to Photon based on the return value of that key's corresponding function in the ComputedInputs object when passed that resolver's parameters at runtime.

**GraphQL Schema Contributions**

The mutation's input type fields with a name that is in the ComputedInputs object are omitted from the GraphQL schema. This modifies one existing input type but does not add new types or remove existing types.

**Example**

```prisma
model User {
  id  Int @id
  name String
  createdWithBrowser String
}
```

```ts
queryType({
  definition(t: any) {
    t.crud.user()
  },
})
mutationType({
  definition(t: any) {
    t.crud.createOneUser({
      computedInputs: {
        createdWithBrowser: ({ root, args, ctx }) => ctx.browser,
      },
    })
  },
})
objectType({
  name: 'User',
  definition: (t: any) => {
    t.model.id()
    t.model.name()
    t.model.createdWithBrowser()
  },
})
```

```gql
type Mutation {
  createOneUser(data: UserCreateInput!): User!
}

type Query {
  user(where: UserWhereUniqueInput!): User
}

type User {
  id: Int!
  name: String!
  createdWithBrowser: String!
}

input UserCreateInput {
  name: String!
}

input UserWhereUniqueInput {
  id: Int
}
```

```gql
mutation createOneUser {
  createOneUser({data: {name: "Benedict"}}) {
    id
    name
    createdWithBrowser
  }
}
```

<br>

### `computedInputs` (global)

```
Record<string, ({root, args, ctx}: MutationResolverParams) => any> | undefined
```

Note: This is an abbreviated version of the ComputedInputs type. The most important thing to undertand each of the object's values will be a function that takes an object with "root", "args" and "ctx" keys that represent the runtime values of the corresponding parameters that are passed to your resolver. For the full type, see [ComputedInputs Type Details](#computedinputs-type-details).

**Applies To**

[`nexusPrismaPlugin()`](#Example)

**About**

Allow clients to omit fields with a given name across all of their GraphQL schema's inputs and infer the value of those fields from context when determining what to pass to Photon

- `undefined` (default) Like `{}`; inputs are unchanged
- `ComputedInputs` (`Record<string, ({root, args, ctx}: MutationResolverParams) => any> | undefined`) [(full type here)](#computedinputs-type-details).

  Keys in the ComputedInputs object will be omitted from all input types. When resolving any mutation at runtime, that mutation's input type will be recursively searched for the omitted keys. Any time one of those keys would have appeared anywhere in the mutation's input type, a value will be passed to Photon based on the return value of that key's corresponding function in the ComputedInputs object when passed the resolver's parameters at runtime.

**GraphQL Schema Contributions**

All input type fields with a name that is in the ComputedInputs object are omitted from the GraphQL schema. This modifies existing input types but does not add new types or remove existing types.

**Example**

```prisma
model User {
  id  Int @id
  name String
  nested Nested[]
  createdWithBrowser String
}

model Nested {
  id Int @id
  name String
  createdWithBrowser String
}
```

```ts
queryType({
  definition(t: any) {
    t.crud.user()
  },
})
mutationType({
  definition(t: any) {
    t.crud.createOneUser()
    t.crud.createOneNested()
  },
})
objectType({
  name: 'User',
  definition: (t: any) => {
    t.model.id()
    t.model.name()
    t.model.nested()
    t.model.createdWithBrowser()
  },
})
objectType({
  name: 'Nested',
  definition: (t: any) => {
    t.model.id()
    t.model.createdWithBrowser()
    t.model.name()
  },
})
```

```ts
nexusPrismaPlugin({
  /*
  Remove fields named "createdWithBrowser" from all mutation input types. When resolving
  a request whose data contains any of these types, the value is inferred from the resvoler's 
  params based on the function we defined below and passed to Photon, even if it's nested.
  This example assumes a Photon context containing a "browser" key that maps to a string
  representing the browser from which the request was made.
  */
  computedInputs: {
    createdWithBrowser: ({ root, args, ctx }) => ctx.browser,
  },
})
```

```gql
type Mutation {
  createOneUser(data: UserCreateInput!): User!
  createOneNested(data: NestedCreateInput!): Nested!
}

type Nested {
  id: Int!
  createdWithBrowser: String!
  name: String!
}

input NestedCreateInput {
  name: String!
  user: UserCreateOneWithoutUserInput
}

input NestedCreateManyWithoutNestedInput {
  create: [NestedCreateWithoutUserInput!]
  connect: [NestedWhereUniqueInput!]
}

input NestedCreateWithoutUserInput {
  name: String!
}

input NestedWhereUniqueInput {
  id: Int
}

type Query {
  user(where: UserWhereUniqueInput!): User
}

type User {
  id: Int!
  name: String!
  nested(skip: Int, after: Int, before: Int, first: Int, last: Int): [Nested!]!
  createdWithBrowser: String!
}

input UserCreateInput {
  name: String!
  nested: NestedCreateManyWithoutNestedInput
}

input UserCreateOneWithoutUserInput {
  create: UserCreateWithoutNestedInput
  connect: UserWhereUniqueInput
}

input UserCreateWithoutNestedInput {
  name: String!
}

input UserWhereUniqueInput {
  id: Int
}
```

```gql
mutation createOneUser {
  createOneUser({data: {name: "Benedict", nested: {name: "Moony"}}) {
    id
    name
    createdWithBrowser
    nested {
      id
      name
      createdWithBrowser
    }
  }
}

mutation createOneNested {
  createOneNested({data: {name: "Moony", user: {connect: {where: {id: 1}}}}}) {
    id
    name
    createdWithBrowser
  }
}
```

If `{user: {connect: {where: {id: 1}}}}` looks familiar, global computedInputs can also be used to determine the user making a request and automatically populate mutations affecting a single user accordingly. For example, assuming Photon's context includes a "userId" key, adding a user key to global computedInputs can simplify the "createOneNested" mutation from the previous example:

```ts
nexusPrismaPlugin({
  ...other config...
  computedInputs: {
    createdWithBrowser: ({ctx}) => ctx.browser,
    user: ({ctx}) => ({ connect: { where: { id: ctx.userId } } }),
  },
})
```

```gql
mutation createOneNested {
  createOneNested({data: {name: "Moony"}}) {
    id
    name
    createdWithBrowser
  }
}
```

<br>

### ComputedInputs Type Details

```ts
/**
 *  Represents arguments required by Photon that will
 *  be derived from a request's input (root, args, and context)
 *  and omitted from the GraphQL API. The object itself maps the
 *  names of these args to a function that takes an object representing
 *  the request's input and returns the value to pass to the photon
 *  arg of the same name.
 */
type ComputedInputs<
  MethodName extends MutationMethodName | null = null
> = Record<string, (params: MutationResolverParams<MethodName>) => any>

type MutationResolverParams<
  MethodName extends MutationMethodName | null = null
> = {
  root: any
  // 'args' will be typed according to its corresponding mutation's input type for resolver-level computedInputs.
  // 'args' will be typed as 'any' for global computedInputs.
  args: MethodName extends null
    ? any
    : MutationResolverArgsParam<
        // Second conditional type is redundant but required due to a TS bug
        MethodName extends null ? never : MethodName
      >
  ctx: Context
}

type MutationResolverArgsParam<
  MethodName extends MutationMethodName
> = core.GetGen<'argTypes'>['Mutation'][MethodName]

type MutationMethodName = keyof core.GetGen<'argTypes'>['Mutation']

type Context = core.GetGen<'context'>
```

ssalbdivad:
@jasonkuhrt Any thoughts about when you might have some bandwidth to take a look at this? For the most part the changes are very consistent with the implementation you requested so while I'm hoping the review process won't be too onerous I'm happy to jump on a call if that would help.

ssalbdivad:
@jasonkuhrt @Weakky Any updates on this?

jasonkuhrt:
Hey @ssalbdivad, going to catch on this tomorrow :)

jasonkuhrt:
Tomorrow slipped, we quietly shipped https://github.com/prisma-labs/graphql-santa (check it out!).

Looping back to this tonight.

jasonkuhrt:
> The only difference is that rather than a single function returning an object, I needed to use an object that maps to individual functions for each key so that we can determine the fields to remove during typegen.

👍🏻

> that mutation's input type will be recursively searched for the omitted keys. Any time one of those keys would have appeared anywhere in the mutation's input type

Feeling a bit of inspiration from CSS selectors suddenly... Imagining there could be a way to increase the specificity, for example from:


```ts
nexusPrismaPlugin({
  computedInputs: {
    user({ ctx }) {
      return { connect: { where: { id: ctx.userId } } }
    },
  },
})
```

To:

```ts
nexusPrismaPlugin({
  computedInputs: {
    user: {
      connect({ ctx }) {
        return { where: { id: ctx.userId } }
      },
    }
  },
})
```

or a string selector syntax

```ts
nexusPrismaPlugin({
  computedInputs: {
    'user.connect': ({ ctx }) => ({ where: { id: ctx.userId } }),
    }
  },
})
```

or a functional DSL

```ts
nexusPrismaPlugin({
  crud: [
    select('user.connect').computeFromContext(ctx => ({ where: { id: ctx.userId } }))
  ],
})
```

But I digress. Just got me thinking. For now, this PR is a fantastic start, but I see that we could generalize _and_ specialize, and go maybe a lot further :)

--- 

Thanks for the docs, it is a great first pass and having written the current monolith completely manually appreciate the labor involved hahah.

The next step for me is to review the diff. I'll dot that over the next few days. Might have a few comments on things to change but definitely biases for merging asap and just following up with changes as new PRs, refactors, whatever it is.

ssalbdivad:
@jasonkuhrt graphql-santa looks very cool from what I've seen so far! In general, most of the refactoring was related to avoiding duplication of my code in builder.ts, but I tried to leave the overall structure of that code about the same. As mentioned, I also made some changes in the dmmf directory to clarify where those types were coming from, but I'm less familiar with internals of prisma in general I may have missed the significance of some of the naming conventions (e.g. internal/external). I'll keep an eye out for your feedback and will address ASAP!

It's obvious just from working with this project for a few days that we've just scratched the surface. I've always been obsessed with DX and the idea that with the right tools, engineers could devote all of their time to problems unique to their teams (hopefully reserving some portion to give back to the  community!). The prisma ecosystem is already one of the most impressive innovations I've seen in getting us closer to that reality. I plan to continue leveraging it wherever possible and looking for opportunities to contribute :bowtie:

README.md
```diff
@@ -2326,6 +2371,336 @@ enum UserStatus {
...
+    id
+    name
+    createdWithBrowser
+  }
+}
+```
+
+<br>
+
+### ComputedInputs Type Details
```
jasonkuhrt:
```suggestion
### `ComputedInputs` Type Details
```

README.md
```diff
@@ -460,6 +465,46 @@ queryType({
...
+    t.crud.createOneUser({
+      computedInputs: {
+        createdByBrowser: ({ root, args, ctx }) => ctx.session.browser,
+      },
+    })
+  },
+})
+```
+
+### Globally Remove a Field from Input Types and Infer its Value
```
jasonkuhrt, line 487: For this case, I think it would be helpful to show:

1. Show some `.crud` that will be impacted by the global config
2. Show some client operations (queries/mutations)
3. Maybe show before vs after, so its very clear the effect the global config is having. In particular that before/after will help new comers get a sense for the feature. That said maybe this point is better for the docs part, while this recipes section is about quick succinct references, optimized for refreshing/skimming, not from-scratch learning.

README.md
```diff
@@ -2326,6 +2371,336 @@ enum UserStatus {
 
 <br>
 
+### `computedInputs` (local)
+
+```
+Record<string, ({root, args, ctx}: MutationResolverParams) => any> | undefined
```
jasonkuhrt: I don't think we need `undefined` type. The property is optional, but don't think we need to make it accept a maybe type.

```suggestion
Record<string, ({root, args, ctx}: MutationResolverParams) => unknown>
```

Also use the `unknown` type instead of `any`.

README.md
```diff
@@ -2326,6 +2371,336 @@ enum UserStatus {
...
+
+**Applies To**
+
+[`t.crud.<mutations>`](#t.crud)
+
+**About**
+
+Allow clients to omit fields from one mutation's corresponding input type and infer the value of those fields from the resolver's params (root, args, and context) at runtime when determining what to pass to Photon.
+
+- `undefined` (default) Like `{}`; inputs are unchanged
```
jasonkuhrt:
```suggestion
```

Per comment above.

README.md
```diff
@@ -2326,6 +2371,336 @@ enum UserStatus {
...
+type ComputedInputs<
+  MethodName extends MutationMethodName | null = null
+> = Record<string, (params: MutationResolverParams<MethodName>) => any>
+
+type MutationResolverParams<
+  MethodName extends MutationMethodName | null = null
+> = {
+  root: any
+  // 'args' will be typed according to its corresponding mutation's input type for resolver-level computedInputs.
+  // 'args' will be typed as 'any' for global computedInputs.
```
jasonkuhrt: We should type it as `unknown`.

README.md
```diff
@@ -2326,6 +2371,336 @@ enum UserStatus {
...
+ *  arg of the same name.
+ */
+type ComputedInputs<
+  MethodName extends MutationMethodName | null = null
+> = Record<string, (params: MutationResolverParams<MethodName>) => any>
+
+type MutationResolverParams<
+  MethodName extends MutationMethodName | null = null
+> = {
+  root: any
```
jasonkuhrt: `undefined`, as `root` always is, which all crud resolvers, being on root types, will be (as opposed to `parent`).

README.md
```diff
@@ -2326,6 +2371,336 @@ enum UserStatus {
...
+  // 'args' will be typed according to its corresponding mutation's input type for resolver-level computedInputs.
+  // 'args' will be typed as 'any' for global computedInputs.
+  args: MethodName extends null
```
jasonkuhrt: What is example where the generic `MethodName` would actually be used?

src/dmmf/DmmfTypes.ts
```diff
@@ -0,0 +1,108 @@
...
+    args: SchemaArg[]
+  }
+  interface InputType {
+    name: string
+    isWhereType?: boolean
+    isOrderType?: boolean
+    atLeastOne?: boolean
+    atMostOne?: boolean
+    fields: SchemaArg[]
+    computedInputs: ComputedInputs
```
jasonkuhrt: ComputedInputs is a concept that sits above DMMF, exclusive to the nexus-prisma api layer. Not sure we want to fuse that into DMMF module which I think represented the concept of the users data layer schema + available photon operations against it.

Whole DMMF module is something I never fully digested and been a "refactor one day thing". So simply not sure about this comment. Thinking out loud.

src/utils.ts
```diff
@@ -1,8 +1,8 @@
-import { isNamedType } from 'graphql'
-import { core } from 'nexus'
 import { relative } from 'path'
 import * as fs from 'fs-extra'
 import * as path from 'path'
+import { core } from 'nexus'
+import { isDeepStrictEqual } from 'util'
```
jasonkuhrt, line 5: TIL

src/utils.ts
```diff
@@ -146,3 +146,40 @@ export function getImportPathRelativeToOutput(
...
+
+/**
+ *  Represents arguments required by Photon that will
+ *  be derived from a request's input (root, args, and context)
+ *  and omitted from the GraphQL API. The object itself maps the
+ *  names of these args to a function that takes an object representing
+ *  the request's input and returns the value to pass to the photon
+ *  arg of the same name.
+ */
+export type ComputedInputs<
```
jasonkuhrt: Note: Left comments about this in the README version.

Also: Lets drop a TODO about autogenerating README docs from this if we're going to maintain an effective copy-paste.

tests/__snapshots__/app.test.ts.snap
```diff
@@ -621,6 +621,41 @@ type GetNexusPrismaInput<
...
+ *  and omitted from the GraphQL API. The object itself maps the
+ *  names of these args to a function that takes an object representing
+ *  the request's input and returns the value to pass to the photon
+ *  arg of the same name.
+ */
+type ComputedInputs<
+  MethodName extends MutationMethodName | null = null
+> = Record<string, (params: MutationResolverParams<MethodName>) => any>
+
+type MutationResolverParams<
```
jasonkuhrt: Thinking about it more I think we might want to keep two sets of typings, the local variant and the global variant. `unknown` is more present at the global level (b/c `args`). Rather than more complicated generics and conditional typing that attempts to unify the two levels, I would explore literal separation. I think this will lead to simpler to understand types.

Sorry for three comments about the same part in three places :P

tests/computedInputs.test.ts
```diff
@@ -0,0 +1,259 @@
+import * as Nexus from 'nexus'
```
jasonkuhrt, line 1: Maybe split computed input tests into global and local suites. Or if keeping as one maybe two describe blocks. In general would prefer different suites, each having no describe nesting, if sensible.

tests/computedInputs.test.ts
```diff
@@ -0,0 +1,259 @@
...
+  const dmmf = await getDmmf(datamodel, {
+    // These are applied globally
+    globallyComputedInputs: { createdWithBrowser: ({ ctx }) => ctx.browser },
+  })
+  expect(
+    addComputedInputs({
+      params: {
+        root: null,
+        // name should be required when creating Nested since the computedInput providing
+        // it is specific to UserCreateInput and therefore shallow
```
jasonkuhrt, line 209: How did you come to the design decision that local computed inputs should be shallow? What are the tradeoffs? No opinion yet myself, just curious.

tests/computedInputs.test.ts
```diff
@@ -0,0 +1,259 @@
...
+  const dmmf = await getDmmf(datamodel, {
+    globallyComputedInputs: {
+      createdWithBrowser: ({ root, args, ctx }) =>
+        `${ctx.browser.slice(1, 2)} ${root} ${args.data.nested.create.name}`,
+    },
+  })
+  expect(
+    addComputedInputs({
+      params: {
+        root: 'Yam',
```
jasonkuhrt: Impossible no? When would `root` have a value?

tests/computedInputs.test.ts
```diff
@@ -0,0 +1,259 @@
...
+      name: 'autopopulated',
+      createdWithBrowser: 'firefox',
+      nested: {
+        create: { createdWithBrowser: 'firefox', name: 'Nested Name' },
+      },
+    },
+  })
+})
+
+it('can use a combination of root, args, and context to compute values', async () => {
```
jasonkuhrt: What about `info`?

tests/computedInputs.test.ts
```diff
@@ -0,0 +1,259 @@
...
+        name: ({ root, args, ctx }) =>
+          `${args.data.nested.create.name} ${ctx.browser.slice(1, 2)} ${root}`,
+      },
+    }),
+  ).toStrictEqual({
+    data: {
+      name: 'Sam i Yam',
+      createdWithBrowser: 'i Yam Sam',
+      nested: {
+        create: { createdWithBrowser: 'i Yam Sam', name: 'Sam' },
```
jasonkuhrt, line 260: Got me thinking about Dr. Seuss now

src/dmmf/transformer.ts
```diff
@@ -66,10 +94,126 @@ function transformArg(arg: DMMF.SchemaArg): ExternalDMMF.SchemaArg {
...
+
+function transformInputType(
+  inputType: DMMF.InputType,
+  globallyComputedInputs: ComputedInputs,
+): DmmfTypes.InputType {
+  const fieldNames = inputType.fields.map(field => field.name)
+  /**
+   * Only global computed inputs are removed during schema transform.
+   * Resolver level computed inputs are filtered as part of the
+   * publishing process. They are then passed to addComputedInputs
```
jasonkuhrt:
```suggestion
   * projecting process. They are then passed to addComputedInputs
```

Term we use now, haven't transitioned the codebase yet.

src/dmmf/transformer.ts
```diff
@@ -66,10 +94,126 @@ function transformArg(arg: DMMF.SchemaArg): ExternalDMMF.SchemaArg {
...
+   * at runtime so their values can be inferred alongside the
+   * global values.
+   */
+  const globallyComputedInputsInType = Object.keys(
+    globallyComputedInputs,
+  ).reduce(
+    (args, key) =>
+      fieldNames.includes(key)
+        ? { ...args, [key]: globallyComputedInputs[key] }
```
jasonkuhrt: fwiw, not sure avoiding a bit of copying here is actually going to matter (nothing in this lib has really been optimized yet):

```suggestion
        ? Object.assign(args, { [key]: globallyComputedInputs[key] }) 
```

src/dmmf/transformer.ts
```diff
@@ -66,10 +94,126 @@ function transformArg(arg: DMMF.SchemaArg): ExternalDMMF.SchemaArg {
...
+      ...addGloballyComputedInputs({
+        inputType,
+        dmmf,
+        params,
+        data: params.args.data,
+      }),
+      ...Object.keys(locallyComputedInputs).reduce(
+        (args, key) => ({
+          ...args,
+          [key]: locallyComputedInputs[key](params),
```
jasonkuhrt: Arguably a bit simpler (no lexical scope look back) with `Object.entries`

```ts
...Object.entries(locallyComputedInputs).reduce((args, [inputFieldName, inputFieldComputer]) => )
```

jasonkuhrt (approved):
@ssalbdivad Nice work. Left some comments. Nothing blocking. I'll merge this in the coming days but if you have time to address the odds and ends I mentioned that'd be great.

I could spend more time reading the builder changes, probably half because I already needed to refresh myself with that module before, but I didn't see anything crazy. Tests are there. I'm more or less satisfied.

🙌

jasonkuhrt:
cc @Weakky

ssalbdivad:
@jasonkuhrt Great, thanks for the feedback! I'll take a look at the comments and update tomorrow.

README.md
ssalbdivad: MethodName is used for local computedInputs to determine the type of the "args" parameter by checking generated Nexus types, hence the type of args:

```ts
export type MutationResolverArgsParam<
  MethodName extends MutationMethodName
> = core.GetGen<'argTypes'>['Mutation'][MethodName]
```

src/dmmf/DmmfTypes.ts
ssalbdivad: Yeah, I found this a bit confusing. I added the type here based on the approach taken for similar options, but I think refactoring it in the future would help clarify. Going to leave it for now as someone with a more holistic vision of the ecosystem would be better qualified.

src/utils.ts
ssalbdivad: Done! I'm a little obsessive with DRY code so it was a bit painful for me to copy this over but I didn't really want to go down that rabbit hole.

tests/computedInputs.test.ts
ssalbdivad: To be honest it was mostly to avoid having to recursively create a new set of types. With the current implementation of the feature, we never create any new types, we just modify existing ones, but that would cause collisions if we allowed deep local computedInputs. Totally open to implementing that now that I have a better understanding of the codebase and especially since that's already how https://github.com/prisma-labs/nexus-prisma/pull/566 works, but would definitely prefer to make that an option in a later PR if possible.

tests/__snapshots__/app.test.ts.snap
ssalbdivad: Split the types up. No problem, I know these diffs are a little insane :stuck_out_tongue:

tests/computedInputs.test.ts
ssalbdivad: I've actually already split these tests up more sensibly as part of https://github.com/prisma-labs/nexus-prisma/pull/566 so I'm going to hold off until that PR. Adding a TODO.

tests/computedInputs.test.ts
ssalbdivad: Yep that makes sense. Given root will never have a value, I've removed it from computedInput's params object.

tests/computedInputs.test.ts
ssalbdivad: I've added info to computedInput's params object!

ssalbdivad:
@jasonkuhrt Implemented most of the changes you suggested and tested end-to-end with our app. Let me know if you have any questions!

tests/computedInputs.test.ts
jasonkuhrt: Yep its a clear limitation that we can address as an isolated feature later on.

jasonkuhrt:
Looking forward to taking one more pass! The only other thing I see left right now I think is having @Weakky take a look (if he wants) and confirming everything is in line with our discussions a few weeks back.

Weakky:
Hey @ssalbdivad, sorry for not participating to the discussion so far, and thanks to @jasonkuhrt for handling it.

I must say, this is incredible work! I'm especially impressed on a technical level that you implemented the globally computed inputs on the DMMF abstraction as a transformer and kept the local computed inputs on a resolver level. 

Code-wise, there's not much for me to add. Everything is clean 💯.

There is just two things I would like to add:

- IIUC, `computedInputs` will/should only compute fields defined in the prisma model. Given that, an improvement we could do in the future is to make `computedInputs` a `Record<PrismaModelFields<TypeName>, ...>` where `PrismaModelFields<TypeName extends string>` is a union string literal of the defined fields. That should provide autocompletion and prevent from inputting wrong field names. Let's keep that for a future improvement, it's fine as it is now for a first iteration!

- Last thing, as a very friendly advice, please please please, create an issue before doing such a huge work in the future. You've been very "lucky" that your worked aligned closely enough to our vision (although it already took you some effort to refactor from your idea to ours), but we could've discarded your work if it didn't align for any reason. This is generally as awkward for us as it is upsetting for the contributor who spent hours working on a super clean PR. Creating an issue makes sure we (and the community) approve the feature and that we can discuss further the API design. If it is ever discarded, you'll save yourself some time and a lot of anger 🙏

Thanks again for the work put into this PR, as much for the code than the detailed explanations throughout the whole PR 🙌

@jasonkuhrt Ready to merge when you are 🎉

Weakky (approved):

ssalbdivad:
@Weakky Thanks for the feedback! This feature ended up evolving organically from some of the frustrations I was having expressing things like "nested creates should always be associated with the same user" without a ton of duplication/boilerplate.  Admittedly, I'm pretty bad at writing code that "feels bad" and moving on, but luckily I have a personal goal to change that anyways so I'll do my best to mock up a spec first in the future :bowtie:

That being said...
Without the benefit of your advice, I stared working on a feature to handle nested relations:
https://github.com/prisma-labs/nexus-prisma/pull/566

I've been consuming this for a few weeks in my app now and the API is much cleaner as a result, but I'm mostly concerned with nailing down an API that will make this as intuitive and convenient as possible. I've held off on doing any more work since I got the basic functionality working, but as soon as I get some feedback on some of the other suggestions in the PR thread with @jasonkuhrt I will take care of it!

tests/__snapshots__/app.test.ts.snap
```diff
@@ -552,6 +552,8 @@ declare global {
 exports[`integrates together: nexus prisma typegen 1`] = `
 "import * as photon from '@prisma/photon';
 import { core } from 'nexus';
+import { GraphQLResolveInfo } from 'graphql';
```
jasonkuhrt, line 555: I think `graphql` can remain a dev dep since we are just importing types from it.

This comment can be addressed in a follow up commit.

ssalbdivad: If someone consuming nexus-prisma didn't have ```graphql``` installed but was using TS, how would ```import { GraphQLResolveInfo } from 'graphql';``` be resolved in the d.ts file? Also ```nexus``` already has ```graphql``` as a peerDependency so effectively ```nexus-prisma``` does as well.

PR #544, opened 2019-11-26
feat: warn on unknown output type

Weakky:


src/hooks.ts
```diff
@@ -0,0 +1,13 @@
+export type OnUnknownFieldName = (info: {
+  unknownFieldName: string
+  error: Error
+  validFieldNames: string[]
+  typeName: string
+}) => void
+
+export type OnUnknownOutputType = (info: {
```
jasonkuhrt:
```suggestion
export type OnUnknownFieldType = (info: {
```

Might make more sense

src/graphql.ts
```diff
@@ -27,3 +27,5 @@ export const scalarsNameValues: ScalarName[] = [
   'ID',
   'Boolean',
 ]
+
+export const isScalarType = (name: string) => scalarsNameValues.includes(name as any)
```
jasonkuhrt:
```suggestion
export const isScalarType = (name: string): boolean => scalarsNameValues.includes(name as any)
```

src/hooks.ts
jasonkuhrt: Other instances of this in the PR would need changing too then.

src/builder.ts
```diff
@@ -335,6 +346,13 @@ export class SchemaBuilder {
           }
           const fieldName = resolvedConfig.alias || field.name
           const type = resolvedConfig.type || field.outputType.type
+
+          if (
+            !this.assertOutputTypeIsDefined(typeName, fieldName, type, stage)
```
jasonkuhrt: Can we use naming to clarify the difference between `typeName` and `type`. e.g. `modelType`? What is it?

jasonkuhrt:
```suggestion
            !this.assertOutputTypeIsDefined(fieldType, fieldName, type, stage)
```

And instances later. Think its clearer to associate that this type name is the type name of the field type.

jasonkuhrt (approved):

src/builder.ts
Weakky: yeah we can! it makes a bit more sense when you see where `type` is going (in the `type` property of `t.field`). Here `type` is the OutputType, and `typeName`, the parent typename in which the dynamic is called. I'll find better names!

PR #551, opened 2019-12-03
feat: more and better error messages

Weakky:
- Better wording for existing errors
- New error when a wrong arg name is passed to filtering or ordering
- Some refactoring

EDIT: Merge of #540 was a bit painful, some testing needs to be done before merging the PR


src/builder.ts
```diff
@@ -565,25 +594,98 @@ export class SchemaBuilder {
       return true
     }
 
-    const message = `${typeName}.${fieldName} is referencing a type "${outputType}" that is not defined in your GraphQL schema.`
+    const message = `Your GraphQL \`${typeName}\` object definition is projecting a field \`${fieldName}\` with \`${outputType}\` as output type, but \`${outputType}\` is not defined in your GraphQL Schema`
```
jasonkuhrt, line 684: The more concrete the better I think. Every mental hope removed in error messages helps. Our code can figure out _which_ kinds of output type are in question.

src/builder.ts
```diff
@@ -565,25 +594,98 @@ export class SchemaBuilder {
...
 
-    if (stage === 'build') {
-      if (isDevMode()) {
```
jasonkuhrt, line 666: Aside: We should investigate how pumpkins build step can do dead code elimination. This includes remove all code behind e.g. `process.env.NODE_ENV !== "production"`. FB already has a tool I'm sure you know for this. We would benefit from adopting similar––in the future.

src/builder.ts
```diff
@@ -565,25 +594,98 @@ export class SchemaBuilder {
...
+        },
+      ],
+      message,
+      stage,
+    )
 
     return false
   }
+
+  assertArgNameExists(
```
jasonkuhrt, line 703: With minor change we can extract this into a pure function? The more methods we accumulate the larger the class-state-mutating surface area becomes. 🤔

src/builder.ts
```diff
@@ -565,25 +594,98 @@ export class SchemaBuilder {
...
+    config: FieldPublisherConfig,
+    stage: 'build' | 'walk',
+    configProperty: 'filtering' | 'ordering',
+  ): { wrongArgNames: string[] } | true {
+    if (!config[configProperty] || config[configProperty] === true) {
+      return true
+    }
+
+    const argNames = Object.keys(
+      config[configProperty] as Record<string, boolean>,
```
jasonkuhrt, line 716: I would prefer `as Index<Boolean>` but I know TimG didn't like that, can't remember what you and I concluded between us in the past about it.

src/hooks.ts
```diff
@@ -11,3 +14,40 @@ export type OnUnknownFieldType = (info: {
...
+  | OnUnknownFieldType
+  | OnUnknownFieldName
+  | OnUnknownArgName
+  | undefined
+
+export function registerHook<T extends Hooks>(
+  hook: T,
+  params: Parameters<Exclude<T, undefined>>,
+  message: string,
+  stage: 'build' | 'walk',
```
jasonkuhrt, line 35: Seems like a good type var to keep in some utils/types module? `type NexusTypeGenStage = ...`. 

Moreover, nexus should export this type. It doesn't?

src/utils.ts
```diff
@@ -120,7 +117,7 @@ export function getImportPathRelativeToOutput(
     return trimIfInNodeModules(to)
   }
 
-  let relativePath = relative(from, to)
+  let relativePath = path.relative(from, to)
```
jasonkuhrt, line 124: 🚲 🏠 

what do you think about

```ts
import * as Path from 'path'
```

vs

```ts
import * as path from 'path'
```

?

tests/hooks.test.ts
```diff
@@ -0,0 +1,246 @@
+import { objectType } from 'nexus'
```
jasonkuhrt, line 1: 👏 (to this module in general)

tests/hooks.test.ts
```diff
@@ -0,0 +1,246 @@
...
+    id  Int @id
+    username String
+  }
+`
+
+  const User = objectType({
+    name: 'User',
+    definition(t: any) {
+      t.model.id()
+      t.model.userName() // wrong projected field
```
jasonkuhrt, line 18: Seems it would be simpler to just do e.g.:

```ts
t.model.wrongProjectedField()
```

or

```ts
t.model.wrong()
```

?

jasonkuhrt (approved):
⛩

Issue #574, opened 2020-01-08
Non-standard ID field names cause errors.


KyleBrost:
I'm running a the following query on an apollo-server instance configured with prisma, nexus-prisma, and MySQL.

```
query {
  user(where : {user_id:"ck50s5bh70001ucvis0k7mesf"}) {
    user_id
    user_email
    user_customer {
      customer_id
    }
  }
}
```

ERROR
```
Invalid `new CustomerClient()` invocation in C:\working\alpha-api\node_modules\@prisma\photon\index.js:196:16

  192 const prefix = this._path.includes('select') ? 'select' : this._path.includes('include') ? 'include' : 'select';
  193 const path = [...this._path, prefix, 'user_customer'];
  194 const newArgs = runtime_1.deepSet(this._args, path, args || true);
  195 this._isList = false;
→ 196 return new CustomerClient(this._dmmf, this._fetcher, this._queryType, this._rootField, this._clientMethod, newArgs, path, this._isList);{
      where: {
        id: undefined,
    ?   user_id?: ID,
    ?   user_email?: String
      },
      select: {
        user_customer: {}
      }
    })

Argument where of type UserWhereUniqueInput needs at least one argument. Available args are listed in green.

Note: Lines with ? are optional.

```

I have attached the various files that could be important.  I'm confused about the error because I thought that nexus-prisma automatically resolved child objects.  Am I mistaken or missing something?

schema.prisma
```
datasource db {
  provider = "mysql"
  url      = "mysql://username:password@host1.example.com/prisma"
}

generator photon {
  provider = "photonjs"
}

model User {
  user_id    String  @default(cuid()) @id
  user_email String  @unique

  user_customer Customer?
  employee Employee?
}

model Employee {
  employee_id String @default(cuid()) @id
  user User?
}

model Customer {
  customer_id String @default(cuid()) @id
  customer_name String
  user User?
}
```

schema.ts
```
import { nexusPrismaPlugin } from 'nexus-prisma'
import { idArg, makeSchema, objectType, stringArg } from 'nexus'


export const User = objectType({
  name: 'User',
  definition(t) {
    t.model.user_id()
    t.model.user_email()
    t.model.user_customer()
    t.model.employee()
  },
})

export const Employee = objectType({
  name: 'Employee',
  definition(t) {
    t.model.employee_id()
    t.model.user()
  },
})

export const Customer = objectType({
  name: 'Customer',
  definition(t) {
    t.model.customer_id()
[...]

PR #590, opened 2020-01-28
fix: support id fields not called id

Weakky:
## Description

Supports id fields that aren't named `id`. It now searches for fields that uses the `@id` directive

## TODO

- Add tests (we have no test infra yet for runtime tests)

## Fixes

- #574


tests/custom_id.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    name: 'Query',
+    definition(t: any) {
+      t.crud.users()
+    },
+  })
+
+  const { graphqlClient, dbClient } = await ctx.getContext({
+    datamodel,
+    types: [Query, User, Post],
+  })
```
Weakky, line 44: Not super happy with the `getContext` naming here but couldn't find anything better

package.json
```diff
@@ -33,21 +33,30 @@
...
     "graphql": "^14.5.4",
+    "graphql-request": "^1.8.2",
+    "graphql-yoga": "^1.18.3",
     "jest": "^24.9.0",
     "jest-watch-typeahead": "^0.4.0",
     "nexus": "^0.12.0-rc.9",
     "prettier": "^1.17.1",
-    "prisma2": "^2.0.0-preview020",
+    "prisma2": "^2.0.0-preview020.3",
+    "react": "^16.12.0",
```
Weakky, line 58: Reason why we need `react`, and `@prisma/cli` is because of `@prisma/lift` which has them as peerDep

src/builder.ts
```diff
@@ -455,6 +455,17 @@ export class SchemaBuilder {
           publisherConfig,
           stage,
         )
+        const mapping = this.dmmf.getMapping(typeName)
```
jasonkuhrt, line 458: Wonder if `mapping` is the best identifier here, or, is there a way to be more specific.

jasonkuhrt (approved):
Brilliant work @Weakky! Your test infrastructure here is super exciting!

PR #618, opened 2020-03-04
fix: ts error when no mutation type is defined

Weakky:


src/typegen.ts
```diff
@@ -317,13 +317,13 @@ export type GlobalMutationResolverParams = BaseMutationResolverParams & {
 }
 
 export type LocalMutationResolverParams<
-  MethodName extends MutationMethodName
+  MethodName extends any
```
Weakky, line 320: it extends `any` here because the rest of the types have generics extending `any` as well and I don't want to update the whole typegen code right now just to fix the initial problem.

Checks are made to make sure `MethodName` extends the available args anyway so nothing should go wrong

jasonkuhrt (approved):
🚂

PR #619, opened 2020-03-04
fix: pin @prisma/client peerDep version

Weakky:
BREAKING CHANGE:
`@prisma/client^2.0.0-preview022` resolves to `2.0.0-preview02x` which can break `nexus-prisma` when new prisma2 releases happen


package.json
```diff
@@ -28,7 +28,7 @@
     "pluralize": "^8.0.0"
   },
   "peerDependencies": {
-    "@prisma/client": "^2.0.0-preview022",
```
jasonkuhrt, line 31: I think we should do same for the `devDependencies` one too, so that we're testing the package with the version that we specify users must have.

jasonkuhrt (approved):

PR #620, opened 2020-03-05
pin the query engine correctly

timsuchanek:


Weakky:
Merging now, thanks a lot Tim 🙏

PR #662, opened 2020-05-06
feat: Add Optional generatedPrismaClientPath path

nnennajohn:
Currently, when the nexusTypes.gen.ts is generated, it uses the prismaClientPath which is written as an absolute url.

If you are doing local development with docker, this poses a problem, as at generation time, the absolute url written is the path inside docker, and shows error on your local path if using a volume mount.

This PR is to take an optional generatedPrismaClientPath that allows the user to pass as a string the relative path to prisma that will be used in the nexusTypes.gen.ts.

This is just an example, and can be merged if you are ok with it. Can also come up with a different solution that aligns more with your conventions. Maybe somehow get the relative

I'm posting this as an alternative to forking this repo to get around this. Hopefully it is considered.

Thank you.

Screen Shot 2020-05-06 at 2 04 22 PM

Without fixing it, when resolving prisma with a custom path, to work with yarn workspaces, It sets the url as an absolute path, and webpack throws an error when trying to resolve it during development mode.

Screen Shot 2020-05-06 at 6 17 37 PM

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -99,6 +99,11 @@ export interface Options {
    */
   prismaClient?: PrismaClientFetcher
   /**
+   * Add optional relate path as string to prisma used in nexusTypes.gen.ts. If not provided,
+   * will use absolute path to prismaClient above.
+   */
+  generatedPrismaClientPath?: string
+  /**
    * Same purpose as for that used in `Nexus.makeSchema`. Follows the same rules
    * and permits the same environment variables. This configuration will completely
    * go away once Nexus has typeGen plugin support.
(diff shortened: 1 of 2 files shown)


jasonkuhrt:
Thanks @nnennajohn, we went with a different implementation. Let us know if this is still an issue for you.

nnennajohn:
HI @jasonkuhrt Thank you. Just saw the updated approach. Thank you. But it does not work for me. Maybe I read the PR wrong, but setting process.env.NEXUS_PRISMA_CLIENT_PATH to a relative path should include the correct path right? But it just shows "prisma" for me.

Thank you.
szoter_annotated_image

jasonkuhrt:
@nnennajohn The setting should be done in code. If that still doesn't work, open a new issue, would seem to be a bug.

PR #674, opened 2020-05-25
feat: support custom resolvers

### LINKED ISSUE #541: Prisma2 CRUD: "before" and "after" hooks
When using the `crud` functionality of nexus-prisma I would like to do something before or after the mutation is executed. 
Maybe this can also be used to manipulate the result, which is returned to the client.

For Example: 
I am using the `t.crud.createOneTicket();` functionality to give the client the ability to create a new ticket. 
After the ticket is created I have to create one or multiple stages, the ticket has to go through based on some calculations.

Right now the client would have to do the calculations and send multiple mutations to be able to do this. Which could lead to a corrupted application state on a flaky network.

It would be great, if the `crud` methods would provide some kind of hooks before and after they are executed. Kind of like how https://github.com/mickhansen/graphql-sequelize#options is doing it:

```
t.crud.createOneTicket({
  before: (root, args, context) => {
    // do something in here
  },
  after: (result, args, context) => {
    // do something in here
  },
});
```

This way the crud operations would become much more flexible and operations which are dependent on each other could be executed with just one network request.

### LINKED ISSUE #684: Add hooks for CRUD nexus prisma fields
Hi everyone :)
Thanks for your amazing job with nexus and prisma. I want today propose a simple workaround for subscriptions and still allowing using autogenerated CRUD (since prisma v2 doesn't propose a built-in solution yet).

For now we can just use crud like this 
```typescript
t.crud.createOneUser()
```

But if you propose hooks, we will be able to add logic before/after the actual CRUD logic. Just like middlewares in fact. The after hook may provide all resolver args and add a new one "result" that is the returned value from the crud operation.
```typescript
t.crud.createOneUser({
  after: (result, parent, args, ctx) => {
    // add more logic
    ctx.pubSub({ //... })
  }
})
```

This will help people like me awaiting for realtime API but also for devs that need to expose CRUD AND add futher logic

craigspaeth:
This is an attempt at addressing [some](https://github.com/graphql-nexus/nexus-schema-plugin-prisma/issues/541#issue-527560086) [issues](https://github.com/graphql-nexus/nexus-schema-plugin-prisma/issues/550#issuecomment-560894721) people are raising for the desire to hook into the CRUD operations. It solves this with middleware that can be used like so... 

````javascript
t.crud.createOneUser({
  middleware: async (root, args, ctx, info, next) => {
    // Something before
    const res = await next()
    // Something after
    return res
  }
})
````

I'm very open to an alternative implementation like `before` and `after` or whatever you all recommend. Whatever shape it takes hooking into the CRUD lifecycle is a feature I'd love to see in this plugin.

This code works for me but two weird things happened when I tried to write a test and clean this up...

1. I got a segfault trying to run tests

```
/bin/sh: line 1: 89527 Segmentation fault: 11  ts-node tests/__ensure-engine
error Command failed with exit code 139.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

2. A bunch of formatting noise got caught in my PR even though I turned off my linter/prettier. Maybe this is because of a git hook or something I'm not seeing?

Let me know what I can do to work past this and get this into a mergeable state.

Update:

Closes #684, #541, #550

### CODE REVIEW

**Comment by Weakky** [comment]
Hey @craigspaeth,

We've taken the liberty to update your PR and get it into a mergeable state. We've also made some modification to your original proposal:

- `middleware` has been renamed to `resolve`. The feature is not a middleware per se as it completely replaces the original resolver and optionally wraps it. There's no stack involved. It also makes the feature integrate seamlessly with the Nexus Schema `resolve` API
- We've made it available on `t.model` as well
- We've renamed `next` to `originalResolve`
- We've updated the typegen so that we get nice type-safety on the `resolve` method for the `root`, `args`, `ctx` params, but also on the same parameters expected by the `originalResolve`r

### Usage

```ts
queryType({
  definition(t) {
    t.crud.users({
      async resolve(root, args, ctx, info, originalResolve) {
        console.log('before')
        const res = await originalResolve(root, args, ctx, info)
        console.log('after')
        return res 
      }
    })
  }
})
```

jasonkuhrt (approved):
🎉

Weakky:
@craigspaeth Thanks a lot for leading that effort, merging now 🚀

craigspaeth:
Wow—amazing, thank you so much @Weakky and team! ❤️ 🚀

Issue #610, opened 2020-02-21
Field alias is not mapping prisma field value


MichalLytek:
Steps to reproduce:

1. Add this line to `examples\blog\src\schema\Blog.ts`:
```diff
  t.model.name()
+ t.model.name({ alias: 'blogName' })
  t.model.createdAt()
```

2. Execute this query:
```graphql
query Blogs {
  blogs {
    id
    createdAt
    name
    blogName
  }
}
```

3. You will receive this response from API:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Blog.blogName.",
      "locations": [
        {
          "line": 6,
          "column": 5
        }
      ],
      "path": [
        "blogs",
        0,
        "blogName"
      ]
    }
  ]
}
```

Without querying for an aliased field, it returns correct data:
```json
{
  "data": {
    "blogs": [
      {
        "id": 1,
        "createdAt": "2020-02-21T09:25:33.583Z",
        "name": "Dynamic Factors Assistant"
      },
      {
        "id": 2,
        "createdAt": "2020-02-21T09:25:33.598Z",
        "name": "Chief Factors Strategist"
      }
    ]
  }
}
```

andriisoldatenko:
@MichalLytek i fixed it using another approach:
```
 t.model.name()
 t.string("blogName", { resolve: ({ name }) => name })
 t.model.createdAt()
```
enjoy!

seans887:
I experienced the exact same issue as @MichalLytek above. I'd prefer to use the built in `alias` property instead of providing a custom resolver if possible, if only so I get the benefit of Prisma's mapping.

divideby0:
I just submitted a PR (#676) which should hopefully address this issue. Awaiting review.

PR #676, opened 2020-05-29
fix: field alias not mapping prisma value

divideby0:
Per #610, Prisma model fields that specify an alias return null in the GraphQL responses. For example:

```typescript
const Customer = objectType({
  type: "Customer",
  definition(t) {
    t.model.full_name({alias: "fullName"})
  }
})
```

The root cause appears to be that the publisher is not providing a resolver in the field config. So when we have a source record from Prisma that looks like this:

```typescript
const prismaSourceObject = {
  "full_name": "Jane Smith"
}
```

Since no custom resolver is specified, the defaultResolver picks it up. The defaultResolver understandably tries extract that value from the source using the alias (e.g. `prismaSourceObject['fullName']`) when it fact it should be extracted from the source via the field name (e.g. `prismaSourceObject['full_name']`). Add a custom resolver appears to fix this.

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -473,15 +473,15 @@ export class SchemaBuilder {
           typeName,
           resolve:
             field.outputType.kind === 'object'
-            ? (root, args, ctx) => {
-                const photon = this.getPhoton(ctx)
-                return photon[lowerFirst(mapping.model)]
-                  ['findOne']({
-                    where: { [idField.name]: root[idField.name] },
-                  })
-                  [field.name](args)
-              }
-            : publisherConfig.alias != field.name
+              ? (root, args, ctx) => {
+                  const photon = this.getPhoton(ctx)
+                  return photon[lowerFirst(mapping.model)]
+                    ['findOne']({
+                      where: { [idField.name]: root[idField.name] },
+                    })
+                    [field.name](args)
+                }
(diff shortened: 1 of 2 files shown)


src/builder.ts
```diff
@@ -473,14 +473,16 @@ export class SchemaBuilder {
...
-                    [field.name](args)
-                }
+            ? (root, args, ctx) => {
+                const photon = this.getPhoton(ctx)
+                return photon[lowerFirst(mapping.model)]
+                  ['findOne']({
+                    where: { [idField.name]: root[idField.name] },
+                  })
+                  [field.name](args)
+              }
```
Akxe: None of these lines changed, consider reversing indentation.

divideby0: Ah that's a good note! I think I had format-on-save checked in vscode and didn't notice I'd shifted everything one column to the left.

divideby0 (commented):
I've applied the suggested indentation reversal. Thank you for the feedback.

Akxe (approved):

Akxe:
I think Prisma should have some eslint/tslint, this seems to happen tad too often.

jasonkuhrt (approved):
Thanks @divideby0!

Merging thanks to your new tests and existing tests passing.

PR #682, opened 2020-06-04
feat: put `t.crud` behind an experimental flag

Weakky:
BREAKING CHANGE: `t.crud` is now disabled by default and needs to be enabled via the `experimentalCRUD` option

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -122,6 +122,13 @@ export interface Options {
      */
     typegen?: string
   }
+  /**
+   * Enable experimental CRUD capabilities.
+   * Add a `t.crud` method in your definition block to generate CRUD resolvers in your `Query` and `Mutation` GraphQL Object Type.
+   *
+   * @default false
+   */
+  experimentalCRUD?: boolean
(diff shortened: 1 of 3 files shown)


jasonkuhrt (approved):
Great, and let’s remember to update docs on nexus repo too.

BjoernRave:
may I ask for the reason of this?

Akxe:
I think the docs were not updated

jasonkuhrt:
@BjoernRave 

> may I ask for the reason of this?

Because there are more ways the `t.crud` API can hit dead ends than not. By comparison the `t.model` API is less prone to such dead ends.

AssisrMatheus:
@jasonkuhrt But should we use t.model in `queryType` too? Is there a shorthand like `t.crud` to expose all queries and mutations?

jasonkuhrt:
@AssisrMatheus no shorthands right now. Follow https://github.com/graphql-nexus/nexus-schema-plugin-prisma/issues/299 for that.

> But should we use t.model in queryType too?

No, instead, if you want, enable crud and use `t.crud`.

Issue #666, opened 2020-05-14
 Error: Input Object type NullableJsonFilter must define one or more fields.


nikolasburk:
I have this Prisma schema:

```prisma
model User {
  id       String   @id
  email    String?  @unique
  name     String
  role     Role?    @default(CUSTOMER)
  jsonData Json?
  posts    Post[]
}

model Post {
  id         String     @id
  createdAt  DateTime   @default(now())
  updatedAt  DateTime   @updatedAt
  content    String?
  published  Boolean    @default(false)
  title      String
  author     User?      @relation(fields: [authorId], references: [id])
  authorId   String?
}

enum Role {
  ADMIN
  CUSTOMER

  @@map("User_role")
}
```

When trying to project relations on my model with `filtering` and `ordering` set to `true`:

```ts
 schema.objectType({
  name: 'User',
  definition(t) {
    t.model.id()
    t.model.email()
    t.model.name()
    t.model.posts({
      filtering: true,
      ordering: true,
    })
  },
})
schema.objectType({
  name: 'Post',
  definition(t) {
    t.model.id()
    t.model.title()
    t.model.content()
    t.model.published()
  },
})
```

I get this error:

```
2134 ● nexus:dev:watcher restarting  --  type: 'change'  file: 'src/generated/nexus.graphql'
 560 ● server listening  --  url: 'http://localhost:4000/'
4294 ✕ app unhandledRejection
       | error  Error: Input Object type NullableJsonFilter must define one or more fields.
       |            at assertValidSchema (/Users/nikolasburk/Desktop/upgrade-guide/prisma-binding/hello-world-mysql/node_modules/graphql/type/validate.js:71:11)
       |            at Object.validate (/Users/nikolasburk/Desktop/upgrade-guide/prisma-binding/hello-world-mysql/node_modules/graphql/validation/validate.js:54:35)
       |            at /Users/nikolasburk/Desktop/upgrade-guide/prisma-binding/hello-world-mysql/node_modules/nexus/dist/runtime/server/handler-graphql.js:28:42
       |            at processTicksAndRejections (internal/process/task_queues.js:89:5)
       |            at async /Users/nikolasburk/Desktop/upgrade-guide/prisma-binding/hello-world-mysql/node_modules/nexus/dist/runtime/server/server.js:87:9
```

Note that everything seems to work fine if I don't include `filtering` and ordering:

```ts
// this works
 schema.objectType({
  name: 'User',
  definition(t) {
[...]

PR #690, opened 2020-06-08
feat: partially support Json type

### LINKED ISSUE #680: Add support for Json type
My schema:
```
model Invoice {
  id                Int                  @default(autoincrement()) @id
  status            InvoiceStatus
  transactions      PaymentTransaction[]
  purchaseDate      DateTime
  details           Json
  products          Product[]
  supportExpiration DateTime
  customerId        Int
  customer          Customer             @relation(fields: [customerId], references: [id])
  sellerId          Int
  seller            Seller               @relation(fields: [sellerId], references: [id])
}
```

Invoice.ts Nexus Prisma Type

```
import { objectType } from '@nexus/schema'

export const Invoice = objectType({
  name: 'Invoice',
  definition(t) {
    t.model.id()
    t.model.details()
  },
})
```

https://prnt.sc/sshpdt
https://prnt.sc/sshq1h

This error should not occur because "Json" is not exactly a type that makes sense to be added to the Nexus Prisma.

Weakky:
This PR fixes #680 and #666

It **does not** add support for JSON filtering as this hasn't been implemented in Prisma yet. It removes the types and args entirely from the schema instead.

### DIFF
--- a/src/publisher.ts
+++ b/src/publisher.ts
@@ -125,6 +125,7 @@ export class Publisher {
         inputType.fields
           .map(field => {
             // TODO: Do not filter JsonFilter once Prisma implements them
+            // https://github.com/prisma/prisma/issues/2563
             if (['JsonFilter', 'NullableJsonFilter'].includes(field.inputType.type)) {
               return null
             }
--- a/src/publisher.ts
+++ b/src/publisher.ts
@@ -125,7 +125,7 @@ export class Publisher {
         inputType.fields
           .map(field => {
             // TODO: Do not filter JsonFilter once Prisma implements them
-            if (field.inputType.type === 'JsonFilter') {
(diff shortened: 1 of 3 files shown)


src/publisher.ts
```diff
@@ -123,21 +123,30 @@ export class Publisher {
...
-                ? // Simply reference the field input type if it's already been visited, otherwise create it
-                  field.inputType.type
-                : this.inputType({
-                    arg: field,
-                    type: this.getTypeFromArg(field),
-                  }),
-            },
-          }))
+          .map(field => {
+            // TODO: Do not filter JsonFilter once Prisma implements them
```
jasonkuhrt, line 127: Is there a gh issue we can pin with the comment?

jasonkuhrt (approved):
👍 👍

src/publisher.ts
Weakky: Just added it! 👍

PR #691, opened 2020-06-08
feat: warn if t.crud is used but not enabled

Weakky:
This PR adds a warning for existing users who used to be able to use `t.crud` without enabling any flag.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,6 @@
 import { plugin } from '@nexus/schema'
 import { build as buildNexusPrismaTypes, Options } from './builder'
+import { colors } from './colors'
 
 /**
  * Create a nexus-prisma plugin to be passed into the Nexus plugins array.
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -36,14 +37,34 @@ import { build as buildNexusPrismaTypes, Options } from './builder'
...
+    onBeforeBuild() {
+      if (wasCrudUsedButDisabled?.() === true) {
+        console.log(`\
+${colors.yellow('WARNING')}: ${colors.green(
+          't.crud',
+        )} ${colors.yellow('is an experimental feature that may evolve and break in the future. It must be explicitely enabled to be used.')}
+Please add ${colors.green(`experimentalCRUD: true`)} in the ${colors.green(
+          'nexusPluginPrisma()',
+        )} if you still wish to enable it.`)
       }
```
Weakky: Core addition of this PR. If you have suggestions to improve the message, I'll happily change it

src/index.ts
```diff
@@ -36,14 +37,34 @@ import { build as buildNexusPrismaTypes, Options } from './builder'
...
-        }),
+        types,
+      }
+    },
+    onBeforeBuild() {
+      if (wasCrudUsedButDisabled?.() === true) {
+        console.log(`\
+${colors.yellow('Warning')}: ${colors.green(
+          't.crud',
+        )} ${colors.yellow('is an experimental feature that may evolve and break in the future. It must be explicitely enabled to be used.')}
```
jasonkuhrt:
```suggestion
        )} ${colors.yellow('is an experimental feature with many practical limitations. You must explicitly enable it before using.')}
```

jasonkuhrt (approved):
👍

PR #692, opened 2020-06-08
feat: support prisma 2

jasonkuhrt:
BREAKING CHANGE:

- Requires Prisma 2.0.0 or higher


package.json
```diff
@@ -56,6 +56,7 @@
     "jest": "25.2.4",
     "jest-watch-typeahead": "0.4.2",
     "node-fetch": "2.6.0",
+    "p-queue": "^6.4.0",
```
jasonkuhrt: This can be removed in latest alpha https://github.com/prisma/prisma/issues/2692

jasonkuhrt:
Some odd things

test suite not exiting correctly

image

funky things with field-resolutions test 

image

package.json
Weakky: Should be removable now that 2.0.0 has been released, right ?

jasonkuhrt: I'll revise this PR to be about 2.0 and yeah let's see.

jasonkuhrt: Ok works.

jasonkuhrt:
https://github.com/prisma/prisma/issues/2703

Weakky (approved):
Looks all good to me 💯

PR #695, opened 2020-06-09
fix: convert nulls to undefined for non-nullable Prisma fields

Weakky:
Take the incoming GraphQL args of a resolver and replaces all `null` values
that maps to a non-nullable field in the Prisma Schema, by `undefined` values.

In Prisma, a `null` value has a specific meaning for the underlying database.
Therefore, `undefined` is used instead to express the optionality of a field.
 
In GraphQL however, no difference is made between `null` and `undefined`.
This is the reason why we need to convert all `null` values that were assigned to `non-nullable` fields to `undefined`.

### DIFF
--- a/src/null.ts
+++ b/src/null.ts
@@ -1,0 +1,46 @@
+import { DmmfDocument, DmmfTypes } from './dmmf'
+
+/**
+ * Take the incoming GraphQL args of a resolver and replaces all `null` values
+ * that maps to a non-nullable field in the Prisma Schema, by `undefined` values.
+ * 
+ * In Prisma, a `null` value has a specific meaning for the underlying database.
+ * Therefore, `undefined` is used instead to express the optionality of a field.
+ * 
+ * In GraphQL however, no difference is made between `null` and `undefined`.
+ * This is the reason why we need to convert all `null` values that were assigned to `non-nullable` fields to `undefined`.
+ */
+export function transformNullsToUndefined(
+  args: Record<string, any>,
+  schemaArgs: Record<string, DmmfTypes.SchemaArg>,
+  dmmf: DmmfDocument,
+) {
+  const keys = Object.keys(args)
+  for (const key of keys) {
+    const val = args[key]
+    const schemaArg = schemaArgs[key]
+
+    if (!schemaArg) {
+      throw new Error(`Could not find schema arg with name: ${key}`)
+    }
+
+    const shouldConvertNullToUndefined =
+      val === null && schemaArg.inputType.isNullable === false
+
+    if (shouldConvertNullToUndefined) {
+      args[key] = undefined
+    } else if (isObject(val)) {
+      const newSchemaArgs = dmmf.getInputTypeWithIndexedFields(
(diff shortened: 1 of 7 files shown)


src/null.ts
```diff
@@ -0,0 +1,46 @@
...
+ * 
+ * In Prisma, a `null` value has a specific meaning for the underlying database.
+ * Therefore, `undefined` is used instead to express the optionality of a field.
+ * 
+ * In GraphQL however, no difference is made between `null` and `undefined`.
+ * This is the reason why we need to convert all `null` values that were assigned to `non-nullable` fields to `undefined`.
+ */
+export function transformNullsToUndefined(
+  args: Record<string, any>,
+  schemaArgs: Record<string, DmmfTypes.SchemaArg>,
```
jasonkuhrt: Could read this as `graphqlArgs` / `prismaArgs`

src/null.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    const shouldConvertNullToUndefined =
+      val === null && schemaArg.inputType.isNullable === false
+
+    if (shouldConvertNullToUndefined) {
+      args[key] = undefined
+    } else if (isObject(val)) {
+      const newSchemaArgs = dmmf.getInputTypeWithIndexedFields(
+        schemaArg.inputType.type,
+      ).fields
```
jasonkuhrt, line 35: I know this is status quo but mentioning as an instance: I find the mix of args and fields terminology confusing.

src/null.ts
```diff
@@ -0,0 +1,46 @@
...
+ * In Prisma, a `null` value has a specific meaning for the underlying database.
+ * Therefore, `undefined` is used instead to express the optionality of a field.
+ * 
+ * In GraphQL however, no difference is made between `null` and `undefined`.
+ * This is the reason why we need to convert all `null` values that were assigned to `non-nullable` fields to `undefined`.
+ */
+export function transformNullsToUndefined(
+  args: Record<string, any>,
+  schemaArgs: Record<string, DmmfTypes.SchemaArg>,
+  dmmf: DmmfDocument,
```
jasonkuhrt, line 16: IIUC `schemaArgs` data is derived from `dmmf`.

Not sure it makes sense in this case, but my inclination would be to have distinctly named and designed dmmf transformations, with sets of functions being tailored to work each one and only one (ideally).

Generic and vague point. 

Might not apply here.

jasonkuhrt (approved):
💯

src/null.ts
Weakky: I get what you mean. The difference is very thin.

src/null.ts
Weakky: Apologies, I think I don't understand your point here

jasonkuhrt: The particular point here being that, there are two params `dmmf` and `schemaArgs` which appear slightly related but slightly not. Was wondering if we can get to a point where there is only 1 param / 1 [maybe large] data structure.

Maybe go deeper in a future call/pr, all good!

PR #696, opened 2020-06-09
fix: support model with a single id field

### LINKED ISSUE #650: Could not find field 'Mutation.undefined' on type Mutation
If I have a model with only one field in my schema and if that field is an autoincremented int, `nexus-prisma` throws an error.

For example, this fails:
```
model User {
  id Int @id @default(autoincrement())
}
```

While this does not:
```
model User {
  id Int @id @default(autoincrement())
  username String @unique
}
```

Fun fact, this does not fail either:
```
model User {
  id String @id @default(uuid())
}
```

The behavior is the same no matter how many models I add.

nexus-prisma version: 0.12.0
@prisma/cli version: 2.0.0-beta.2
@prisma/client version: 2.0.0-beta.2
@nexus/schema version: 0.13.1

Edit: This bug also occurs when the schema.prisma is empty

Weakky:
Closes #650


tests/schema/__snapshots__/crud.test.ts.snap
```diff
@@ -0,0 +1,317 @@
...
+  }
+}
+
+interface NexusPrismaTypes {
+  Query: {
+    user: 'User'
+    users: 'User'
+
+  },
+  Mutation: {
```
jasonkuhrt, line 289: For example there is no upsert?

jasonkuhrt (approved):
👍

tests/schema/__snapshots__/crud.test.ts.snap
Weakky: Yup exactly!

PR #699, opened 2020-06-09
feat: revert to old pagination model

Weakky:
This PR brings the concept of "Pagination Strategies", allowing us to customize how pagination is handled.

It also brings back the old pagination model so that the new Prisma Client pagination doesn't break existing users.

### EDIT

This PR is partially breaking the old pagination.

- `skip` has been removed to move towards a cursor-based pagination style only. It'll be possible to bring it back in case lots of people were using the offset-based pagination.

- Also, a few combinations aren't possible to handle anymore such as:
  - `after` + `last`
  - `before` + `first`

## TODO
- [x] Validate error handling for the pagination args
- [x] Validate mapping between pagination args to PC args

### DIFF
--- a/src/pagination.ts
+++ b/src/pagination.ts
@@ -10,13 +10,13 @@ interface PaginationResult {
 }
 
 export interface PaginationStrategy<T = object> {
-  transformDmmfArgs: (params: {
+  paginationArgNames: string[]
+  transformDmmfArgs(params: {
     args: DmmfTypes.SchemaArg[]
     paginationArgNames: string[]
     field: DMMF.SchemaField
-  }) => DmmfTypes.SchemaArg[]
-  paginationArgNames: string[]
-  resolve: (args: T) => PaginationResult
+  }): DmmfTypes.SchemaArg[]
+  resolve(args: T): PaginationResult
 }
 
 const relayLikePaginationArgs: Record<
--- a/src/pagination.ts
+++ b/src/pagination.ts
@@ -1,7 +1,7 @@
 import { DmmfTypes, getReturnTypeName } from './dmmf'
 import { DMMF } from '@prisma/client/runtime'
 
-export interface PaginationStrategy<T = any> {
+export interface PaginationStrategy<T = object> {
   transformDmmfArgs: (params: {
     args: DmmfTypes.SchemaArg[]
     paginationArgNames: string[]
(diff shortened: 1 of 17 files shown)


src/typegen.ts
```diff
@@ -216,14 +225,23 @@ function renderNexusPrismaMethods(dmmf: DmmfDocument) {
...
+  .map(m => `  ${m.name}: NexusPrismaFields<'${m.name}'>`)
   .join('\n')}
   Query: NexusPrismaFields<'Query'>
   Mutation: NexusPrismaFields<'Mutation'>
 }
   `
 }
 
+function renderPaginationType(paginationStrategy: PaginationStrategy) {
+  return `\
```
jasonkuhrt: We have the `outdent` dep we can use now.

src/typegen.ts
```diff
@@ -25,15 +27,23 @@ export function doGenerate(
...
   } else {
     return hardWriteFile(options.typegenPath, tsDeclaration)
   }
 }
 
-export function render(dmmf: DmmfDocument, prismaClientPath: string) {
+export function render(
+  dmmf: DmmfDocument,
+  prismaClientPath: string,
+  paginationStrategy: PaginationStrategy,
```
jasonkuhrt: time for keyword arguments I think

src/pagination.ts
```diff
@@ -0,0 +1,174 @@
...
+import { DMMF } from '@prisma/client/runtime'
+
+export interface PaginationStrategy<T = object> {
+  transformDmmfArgs: (params: {
+    args: DmmfTypes.SchemaArg[]
+    paginationArgNames: string[]
+    field: DMMF.SchemaField
+  }) => DmmfTypes.SchemaArg[]
+  paginationArgNames: string[]
+  resolve: (
```
jasonkuhrt: A dx benefit of `resolve(...` style typings is autocomplete shows it as a method. The point is it shows different than a property. It ultimately can help separate callables from non-callabs, basically.

src/pagination.ts
```diff
@@ -0,0 +1,174 @@
+import { DmmfTypes, getReturnTypeName } from './dmmf'
```
jasonkuhrt, line 1: Maybe a layout like:

```
pagination/
	core.ts                      <-- or strategy.ts  ?
	strategyRelayLike.ts
```

?

jasonkuhrt (approved):
A superficial pass, don't deeply understand, but overall impression is 💯.

src/pagination.ts
Weakky: I thought about it and intentionally didn't want with such a layout as I preferred waiting for the day we'd have multiple pagination strategy

PR #704, opened 2020-06-11
feat(typegen): relative @prisma/client import when path given

jasonkuhrt:
close #662


Weakky (approved):
💯

src/typegen.ts
```diff
@@ -26,12 +27,19 @@ export function doGenerate(
   sync: boolean,
   options: Options,
 ): void | Promise<void> {
+  const paginationStrategy = options.paginationStrategy
+  const prismaClientImportId =
+    Path.isAbsolute(options.typegenPath) &&
+    Path.isAbsolute(options.prismaClientPath)
+      ? Path.relative(options.typegenPath, options.prismaClientPath)
```
jasonkuhrt: This is the crux of the solution.

I'm conservative to only apply the relative solution when both inputs are absolute.

nnennajohn:
Hi @jasonkuhrt Reading through this pr, I tried setting NEXUS_PRISMA_CLIENT_PATH, but the generated file still just reference prisma. Is there a doc on how to pass relative path properly?

Thank you.

szoter_annotated_image

PR #709, opened 2020-06-13
feat(pagination): configurable pagination, prisma option

Gomah:
Following this discussion: https://github.com/graphql-nexus/nexus-schema-plugin-prisma/issues/677#issuecomment-643279257, I think it would be good to have the possibility to use the native-like pagination introduced in Prisma beta 7 as it's confusing (at least for me) to use different args between the schema & the query engine.

What's in this PR:

- Add supports for "native" like pagination
- Use Relay Like pagination by default
- Pagination tests (relay & native)


src/pagination.ts
```diff
@@ -151,6 +199,28 @@ export const relayLikePaginationStrategy: PaginationStrategy<RelayLikePagination
 
     return newArgs
   },
+  },
+  native: {
+     paginationArgNames: Object.keys(nativePaginationArgs),
+  transformDmmfArgs({ args, paginationArgNames, field }) {
```
Gomah: I assume I can just return `args` there as there's nothing to transform?

jasonkuhrt (approved):
Thanks @Gomah! We took the liberty of tweaking a few things so that its ready as soon as it hits trunk.

Gomah:
Cheers @jasonkuhrt ✌️

macrozone:
nice! a bit disappointed that relay style is the default and not the `prisma` style as probably most want to have the same semantics on the graphql query as on prisma. Also prisma style is strictly superior as it supports offset based pagination AND cursor based (even combination). We have now a pitfall in new projects, because you might forget to set this option.

Weakky:
We understand your point @macrozone. Our vision is that `nexus-prisma` should generate an API that follows best GraphQL practices, and Prisma's pagination strategy isn't idiomatic to GraphQL at all.

Also, in most cases, cursor-based pagination is superior to the offset-based one, in which case the relay-like pagination fits perfectly and is something that most people are already used to.

Finally, sticking to the relay-like pagination as default prevent from breaking every existing project. It's still breaking as we've removed the `skip` argument in the `relay` pagination strategy, but if you were using the cursor-based pagination, you should be fine.

jasonkuhrt:
@macrozone feel free to create a new feature request to make your proposal and discuss further if you feel strongly about it.

The default will have to align with pagination at the Nexus level since a default experience wherein a Nexus app would generate different pagination between `t.crud` and non `t.crud` (e.g. `t.connection` today, maybe `t.paginated` tomorrow) is unacceptable we think.

In other words we think a user should have to go _out of their way_ to create a GraphQL API where pagination styles are inconsistent.

This is another reason we did not default to prisma strategy today; in Nexus, outside `t.crud`, there is no standard way to use a "prisma" pagination strategy.

Anything can change though, if the community doesn't want relay pagination as default, it won't be. Still too early to tell though.

PR #712, opened 2020-06-15
feat: enable configuring scalars

Weakky:


tests/runtime/scalars.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  try {
+    await graphqlClient.request(`{
+      users {
+        createdAt
+      }
+    }`)
+  } catch (e) {
+    expect(e.message).toContain('not a date')
+  }
+})
```
jasonkuhrt, line 56: Should we add a test with a custom resolver show that the custom scalar data coming through in the args of the resolver is not just a string? And that it can be forward to prisma? Use `graphql-scalars` lib as a dev dep to import something real? Or just roll our own in the test?

tests/__utils.ts
```diff
@@ -51,13 +51,15 @@ export async function generateSchemaAndTypes(
   types: any[],
   options?: TransformOptions & {
     experimentalCRUD?: boolean
-    plugins?: Nexus.core.NexusPlugin[]
+    plugins?: Nexus.core.NexusPlugin[],
+    scalars?: Record<string, GQL.GraphQLScalarType>
```
jasonkuhrt, line 54: Shouldn't this be not an unbounded indexed type but an object for scalars mapping to the set of support prisma scalars? e.g. `{ json: ..., date: ... }`. My reference was this: https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema/data-model#scalar-types.

Weakky: Done in c3be2a5 and c9c2a86

tests/runtime/scalars.test.ts
Weakky: Done ✅

src/builder.ts
```diff
@@ -143,12 +143,21 @@ export interface Options {
...
+   * @default {}
+   */
+  scalars?: Partial<Record<Typegen.GetGen<'scalars'>, GraphQLScalarType>>
   computedInputs?: GlobalComputedInputs
 }
 
 export interface InternalOptions extends Options {
   dmmf?: DmmfDocument // For testing
   nexusBuilder: Nexus.PluginBuilderLens
+  nexusPrismaImportId?: string,
```
Weakky, line 160: Same stuff as for `nexus`. If the lib is not hoisted, the import to `nexus-prisma/typegen` in the typegen file will fail, so we'll probably need to re-expose that on nexus and change the `importId` to something like `nexus-plugin-prisma/schema/typegen`. Could be `nexus-plugin-prisma/typegen`.

src/typegen/render.ts
```diff
@@ -0,0 +1,257 @@
...
+}
+
+export function render(params: {
+  dmmf: DmmfDocument
+  prismaClientImportId: string
+  nexusPrismaImportId?: string
+  paginationStrategy: PaginationStrategy
+}) {
+  return `\
+import * as Typegen from '${params.nexusPrismaImportId ?? 'nexus-prisma/typegen'}'
```
Weakky, line 55: All the static stuff is now statically imported from the library itself instead of being re-generated every time. See comment above about some of the implications

src/typegen/render.ts
```diff
@@ -0,0 +1,257 @@
...
+${renderNexusPrismaMethods(params.dmmf)}
+
+interface NexusPrismaGenTypes {
+  inputs: NexusPrismaInputs
+  outputs: NexusPrismaOutputs
+  methods: NexusPrismaMethods
+  models: PrismaModels
+  pagination: Pagination
+  scalars: CustomScalars
+}
```
Weakky, line 83: Same pattern as Nexus Schema

jasonkuhrt (approved):
💯

PR #731, opened 2020-06-30
feat: postinstall warning when user has prisma deps

jasonkuhrt:
closes graphql-nexus/nexus#1078

image

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,5 +1,5 @@
 const chalk = require('chalk')
-const jetpack = require('fs-jetpack')
+const fs = require('fs-jetpack')
 const path = require('path')
 
 if (process.env.INIT_CWD) {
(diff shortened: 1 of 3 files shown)


scripts/postinstall.js
```diff
@@ -1,5 +1,4 @@
-const chalk = require('chalk')
-const jetpack = require('fs-jetpack')
+const fs = require('fs-jetpack')
```
jasonkuhrt, line 1: OT

Started to get a bit worried about this being fragile. If not hoisted, we're going to get a runtime error.

scripts/postinstall.js
```diff
@@ -9,23 +8,14 @@ if (process.env.INIT_CWD) {
...
+if (fs.exists(from)) {
+  const destDir = path.resolve(nodeModulesFolder, '@types/nexus-plugin-prisma')
 
-  jetpack.dir(destDir)
-  jetpack.copy(from, path.join(destDir, 'index.d.ts'), { overwrite: true })
+  fs.dir(destDir)
+  fs.copy(from, path.join(destDir, 'index.d.ts'), { overwrite: true })
 }
 
-console.log(chalk.bold.yellowBright('----------------------------------'))
```
jasonkuhrt, line 22: Don't think this is very helpful anymore and even if we wanted to keep it, would need rethinking now that the packages are merged.

No?

scripts/prisma-deps-check.js
```diff
@@ -0,0 +1,86 @@
...
+
+const devDeps = pj.devDependencies || []
+foundPrismaDeps.push(
+  ...Object.keys(devDeps)
+    .filter(isPrismaDep)
+    // dedupe
+    .filter((name) => !foundPrismaDeps.includes(name))
+)
+
+const message = `
```
jasonkuhrt, line 28: I didn't go with the box design because of variable width content.

PR #735, opened 2020-06-30
improve: remove @nexus/schema & graphql from peer deps

Weakky:
Rendering:
image


package.json
```diff
@@ -22,26 +22,22 @@
     "release:stable": "dripip stable",
     "release:pr": "dripip pr",
     "clean": "rm -rf dist",
-    "build": "ts-node scripts/ensure-prisma2-version && yarn -s clean && yarn -s build:module-facades && tsc && ts-node scripts/post-build",
+    "build": "yarn -s clean && yarn -s build:module-facades && tsc && ts-node scripts/post-build",
```
jasonkuhrt:
```suggestion
    "build": "yarn clean && yarn build:module-facades && tsc && ts-node scripts/post-build",
```

Recently learned yarn is smart enough to do this automatically within scripts

scripts/ensure-prisma2-version.ts
```diff
@@ -1,26 +0,0 @@
-/**
```
jasonkuhrt, line 1: Why not needed anymore?

src/schema/utils.ts
```diff
@@ -185,3 +186,22 @@ export const isEmptyObject = (o: any) => isDeepStrictEqual(o, {})
...
+export function ensureDepIsInstalled(depName: string) {
+  try {
+    require(depName)
+  } catch (err) {
+    if (err.code === 'MODULE_NOT_FOUND') {
+      console.error(
+        `${chalk.redBright('ERROR:')} ${chalk.greenBright(
+          depName
+        )} must be installed for nexus-plugin-prisma to work properly. Please run \`${chalk.greenBright(
+          `npm install ${depName}`
```
jasonkuhrt, line 199: Suggestion:

> When nexus-plugin-prisma is used as a Nexus schema plugin then you must have ... installed.

Point: Can we lower confusion by qualifying this is just for schema plugin use-case?

jasonkuhrt (approved):

src/schema/utils.ts
```diff
@@ -185,3 +186,22 @@ export const isEmptyObject = (o: any) => isDeepStrictEqual(o, {})
 export function keys<A extends object>(a: A): (keyof A)[] {
   return Object.keys(a) as any
 }
+
+export function ensureDepIsInstalled(depName: string) {
+  try {
+    require(depName)
```
jasonkuhrt, line 192: We could make this check against the package.json contents.

Doing so would be stricter.

Not sure about the benefit...

It would for example make it not nice to use prisma nexus schema plugin in a framework context because the framework [will] complains when you instal graphql or @nexus/schema deps. Catch 22.

More than ok with how you did this though! Just mentioning in case helpful

scripts/ensure-prisma2-version.ts
Weakky: Because this script was previously there to ensure that dev deps and defined peer deps were in sync. That was mainly to prevent us from forgetting to update the peer deps when upgrading prisma.
Now that there's no longer peer deps at all, the check is no longer needed.

package.json
Weakky: That's cool, updated 👍

src/schema/utils.ts
Weakky: Yeah, interesting. Your approach would ensure that bad stuff cannot happen in prod. eg: you have graphql or nexus schema installed as dev deps locally, but it fails in prod as the dev deps aren't installed.

I don't feel comfortable doing IOs to find the package.json in the schema lib though. We'll have to recursively walk through the files again and it sucks to have to pay that price in prod

jasonkuhrt: Good point, how about:

```ts
if (process.env.NODE_ENV === 'production'){ /* cheap check */ } else { /* expensive check */ }
```

About my point about schema plugin not working in fw context, easy to make exception for that case in our check. If we want to, someday.

src/schema/utils.ts
Weakky: image

All good ?

src/schema/utils.ts
Weakky: Do you mind if we wait for the bad situations described above to happen? I'm worried about the package.json walk, even in dev. We know from the framework the kind of issues it can create to look for files, in some situations like monorepos and stuff like that. I'd like to not create a whole new category of bugs because of it. If you strongly think we should do it then I'll make the change, but my feeling is that staying simpler here is better, especially for something that might be a hard blocker

jasonkuhrt: ~Could use https://github.com/sindresorhus/read-pkg-up~

jasonkuhrt: Seeing your comment now, yep sounds good 👍

nnennajohn:
Hi @Weakky , do you know if this tested with Yarn 2 or if its broken? I get an error that @nexus/schema is not installed, even though it is. 

Thank you.

Screen Shot 2020-07-01 at 1 16 52 AM

jasonkuhrt:
We've never tested with Yarn 2. Does it work with Yarn? If not, please open a new issue, thanks!

PR #743, opened 2020-07-08
fix: different crud warning if FW or schema plugin

### LINKED ISSUE #728: Typo in a Warning message. Should be `nexusPluginPrisma` instead of `nexusPrismaPlugin`
## Bug description

I had a project I made in Prisma 5-6 months ago & haven’t touched since. I was just trying to update it to Prisma Stable. There is a warning that says:

> Please add experimentalCRUD: true in the nexusPluginPrisma() constructor if you still wish to enable it.

But instead of `nexusPluginPrisma` it should be `nexusPrismaPlugin` I think.

My code is as follows that is working:

```js
const { nexusPrismaPlugin } = require('nexus-prisma')

const schema = makeSchema({
  plugins: [
    nexusPrismaPlugin({
      experimentalCRUD: true,
    }),
  ],
  ...
})
```

Also, curious if this is the way to go or I have to use https://www.nexusjs.org/#/plugins/prisma in the future?

Weakky:
closes https://github.com/graphql-nexus/nexus/issues/1146
closes https://github.com/graphql-nexus/nexus/issues/1063
closes #728

### DIFF
--- a/src/framework/runtime.ts
+++ b/src/framework/runtime.ts
@@ -2,15 +2,13 @@ import chalk from 'chalk'
 import { RuntimePlugin } from 'nexus/plugin'
 import * as Path from 'path'
 import { nexusSchemaPrisma } from '../schema'
-import { InternalOptions } from '../schema/builder'
+import { InternalPublicOptions } from '../schema/builder'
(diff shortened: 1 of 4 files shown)


jasonkuhrt (approved):
🧯

Issue #749, opened 2020-07-16
Pagination parameter is required when nonNullDefaults is true


nimashoghi:
I have been running into an issue where the `skip` and `take` parameters of a paginated query would be marked as required is `nonNullDefaults` is true.

You can reproduce it by creating any paginated query (either using `crud` or `model`) when `nonNullDefaults` is true for inputs.

#750 fixes this for me.

PR #750, opened 2020-07-16
Fixed issue with pagination parameter nullability when nonNullDefaults is true

nimashoghi:
See issue #749


nafkhanzam:
bump, this really needs to be merged

PR #796, opened 2020-08-17
fix: Fix relation after deletion

### LINKED ISSUE #851: Can't return associated records of deleted record
#### Nexus Report
```json
{
  "node": "v12.18.3",
  "os": {
    "platform": "linux",
    "release": "5.8.5"
  },
  "nexus": "^0.27.0-next.5",
  "plugins": [
    "nexus-plugin-prisma"
  ],
  "otherDependencies": {},
  "devDependencies": {},
  "hasAppModule": true,
  "packageManager": "npm",
  "errorsWhileGatheringReport": {
    "gettingLayout": null,
    "gettingPluginManifests": null
  }
}
```

#### Description
When deleting a record which has associated records nexus fails to return the deleted data if you request the associated record.

The actual problem is that there is a request for the data
```sql
SELECT `test`.`Round`.`id`, `test`.`Round`.`studentId` FROM `test`.`Round` WHERE `test`.`Round`.`id` = ? LIMIT ? OFFSET ?
```
after the data had been deleted.

#### Repro
The [README.md](https://github.com/mohe2015/nexus-bug1/blob/main/README.md) should contain all necessary steps.
https://github.com/mohe2015/nexus-bug1/

Btw this is a blocker for me so if there is a workaround I would love to hear about it.

mohe2015:
May fix https://github.com/graphql-nexus/nexus/issues/1343

It's really late here so this fix may not have the indended quality. Also it's really hard to understand the code for me.

### DIFF
--- a/src/schema/builder.ts
+++ b/src/schema/builder.ts
@@ -343,6 +343,7 @@ export class SchemaBuilder {
 
               args = this.paginationStrategy.resolve(args)
 
+              console.log("346 prismaClient")
               return prismaClient[mappedField.prismaClientAccessor][mappedField.operation](args)
             }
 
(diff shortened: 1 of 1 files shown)


mohe2015:
Seems fixed or I'm not depending on it any more.

Issue #763, opened 2020-07-23
rename `prisma2` occurrences with `prisma`


janpio:
These should all be replaced with `prisma`: https://github.com/graphql-nexus/nexus-plugin-prisma/search?q=prisma2&unscoped_q=prisma2

PR #803, opened 2020-08-25
typo: prisma2 to prisma

alanstriglio:
`prisma2` occurrences were replaced to `prisma`. 

#763


jasonkuhrt:
Since a large rebase is needed here I'll just take care of this. Thanks though @alanstriglio!

Issue #537, opened 2019-11-19
TS2300: Duplicate identifier


schrobak:
I have this `schema.prisma`:

```gql
model Media {
  id String @default(cuid()) @id @unique
}

model Audio {
  id String @default(cuid()) @id @unique
}
```

After generating the supporting type of nexus prisma I end up with 
```ts
interface NexusPrismaTypes {
  Query: {
    media: "Media";
    media: "Media";
    audio: "Audio";
    audio: "Audio";
  }
}
```

Any idea why those words are not changed to their equivalent plural form (event if they will be grammatically incorrect)?

schrobak:
I actually find out that `pluralize` was causing the issue here. The solution for me was to add following line in the schema generation script:
```
import * as pluralize from "pluralize";

pluralize.addIrregularRule("media", "medias");
pluralize.addIrregularRule("audio", "audios");
```

WellFail:
+1

Devlin556:
+1

denisRabov:
+1

WellFail:
Similar problem with numbers in model names:
```
    post3: 'Post3'
    post3S: 'Post3'
``` 
Currently, last character 's' is generated in the upper case.

jasonkuhrt:
Thanks for reporting this. Looks like we need to review our pluralize dep. it really ought to handle pretty much any case I think. Irregular shouldn’t default to doing nothing. I’m on mobile now but will check later if we can just better use the pluralize lib e.g. config.

jonestristand:
Another example. Model name is staff, so plural is also staff. Then cannot access t.crud.staff for both the list and individual variants of the model

wbunting:
Another thing you can do if your underlying database table name does not have a valid "pluralized" version then you can use the `@@map()` to point a model at that table and as long as the model name can be pluralized it should work as expected

sfratini:
Also wanted to add this does not work:
```
pluralize.addIrregularRule("categories", "categoriess");
```
I am aware prisma recommends to use single instead of plural on the db model, but still, this breaks current deployment. I am on stage where I can rename tables so I am fine, but will not always be the case.

alanstriglio:
Hello, 
since Prisma (Prisma 2) doesn't use anymore pluralize for querying Models, I think the best way to solve this is by using same methods on Prisma: 

```js
prisma.user.findOne()
prisma.user.findMany()
prisma.user.createOne()

// etc

[...]

PR #804, opened 2020-08-25
Fix/remove crud pluralize

alanstriglio:
This branch fix the pluralize issue #537 

Instead of **Query** `t.crud.user` and `t.crud.users` the new methods are now `t.crud.findOneUser` and `t.crud.findManyUser` following the Mutation methods style `createOneUser`.

> I know this is a great change for current user, so take your time to evaluate this PR.

Also I fix some tests that was broke/broken.


P4sca1:
What do you think about using pluralization after `findMany`? So `findManyUser` becomes `findManyUsers`?

alanstriglio:
That approach would solve the problem just as well, but it wouldn't be in line with the prisma.io project.

nealoke:
Any idea when this will be merged 🙏?

PR #805, opened 2020-08-26
feat: upgrade to latest prisma 2.6-dev

### LINKED ISSUE #798: Upgrade to prisma 2.5.0
Prisma just released `2.5.0` with new features. It would be nice to bump to the newest version

jasonkuhrt:
closes #798

We had some issues with the tests with 2.5.1. No one could readily figure out the issue. We discovered that the issue we were having does not happen with current dev. Current dev will become 2.6. We decided to release a canary with current dev. Next stable of prisma plugin will be cut when 2.6 is cut. Next week.

#### TODO

- [x] start a new pr on the docs with changes reflecting the changes in this pr
  https://github.com/graphql-nexus/nexus/pull/1385
- [x] check order by is still working at runtime
- [x] ~check that we don't have any bad `isList` usage in codebase. Compare to the isList change in dmmf~
- [x] ~collapse nested filters back into just the normal/existing filters~
- [x] craft the breaking change descriptions; here, then commit msg

BREAKING CHANGES:

- `UUID` scalar has been removed. Thus so has its filter types e.g. `UUIDFilter`. Prisma models like this now result in GraphQL `String` types. This behaviour of `uuid` now aligns with `cuid`. Note that in the future we would like both of these cases to map to `ID` GraphQL types ([issue](https://github.com/graphql-nexus/nexus-plugin-prisma/issues/807)).

  ```prisma
  model User {
    id String @id @default(uuid()) # now maps to String type in GraphQL schema (used to be UUID)
  }
  ```

- There are new nested scalar filter types. Scalar filter types have a `not` field. Its type has changed from the type of respective Scalar to the type of the respective nested filter. Below you can see an example, how the string filter type has changed. The pattern here is the same for all filter types. This change prepares for eventually supporting the [new Prisma Client `mode` option](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/filtering#case-insensitive-filtering) on string filters to control [case sensitivity](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/case-sensitivity). The top level filter types will eventually have the `mode` option. The `NestedStringFilter` type however will not. Achieving that separation is why two types are necessary. All scalar filters have this type separation pattern now. But only String filters are currently planned to make use of it. That is, for example, Int filters will not have a `mode`. We have decided however not to collapse the types where possible. Instead we have opted for consistentcy. In the future we may make an option to generate a GraphQL schema where scalar filters are collapsed where possible. In sum here are the affected scalar types:

  ```
  StringFilter
  StringNullableFilter
  IntFilter
  IntNullableFilter
  FloatFilter
  FloatNullableFilter
  DateTimeFilter
  DateTimeNullableFilter
  BooleanFilter
  BooleanNullableFilter
  ```

  ```diff
    input StringFilter {
      contains: String
      endsWith: String
      equals: String
      gt: String
      gte: String
      in: [String!]
      lt: String
      lte: String
  -   not: String
  +   not: NestedStringFilter
      notIn: [String!]
      startsWith: String
    }

  +  input NestedStringFilter {
  +    contains: String
  +    endsWith: String
  +    equals: String
  +    gt: String
  +    gte: String
  +    in: [String!]
  +    lt: String
  +    lte: String
  +    not: NestedStringFilter
  +    notIn: [String!]
  +    startsWith: String
  +  }
  ```

- The relation filter type naming pattern has changed from `RM_Filter` to `RM_ListRelationFilter`. As a reminder relation filter types show up in the following places (below, the `RM_Filter` type is what we're interested in):

  ```gql
  query {
    # When filtering option is enabled
    Ms(where: M_WhereInput, ...): [M!]!
  }
  mutation {
    updateMany_M(where: M_WhereInput, ...) BatchPayload!
    deleteMany_M(where: M_WhereInput): BatchPayload!
  }

  input RM_Filter { # the name pattern of this type changes to RM_ListRelationFilter
    every: RM_WhereInput
    none: RM_WhereInput
    some: RM_WhereInput
  }
  ```

- The ordering input fields have changed to be a list of order by types rather than just one. This ensures stable ordering where there sometimes can not be for object keys ([details](https://github.com/prisma/prisma/releases/tag/2.4.1)). As a reminder ordering input fields show up in `t.crud.<BatchRead>` and `t.model.<ListRelation>` locations. For details see https://nexusjs.org/pluginss/prisma/runtime#ordering. Here is how the schema contribution changes:

  ```diff
  # t.crud.<BatchRead>
  - M(orderBy: M_OrderByInput)
  + M(orderBy: [M_OrderByInput!])
  # t.model.<ListRelation>
  type M {
  -  MF(orderBy: M_OrderByInput)
  +  MF(orderBy: [M_OrderByInput!])
  }
  ```


tests/schema/__app/generated/schema.graphql
```diff
@@ -9,7 +9,7 @@ type BatchPayload {
 type Bubble {
   createdAt: DateTime!
   id: String!
-  members(after: UserWhereUniqueInput, orderBy: BubbleMembersOrderByInput, where: BubbleMembersWhereInput): [User!]!
+  members(after: UserWhereUniqueInput, orderBy: [BubbleMembersOrderByInput!], where: BubbleMembersWhereInput): [User!]!
```
jasonkuhrt, line 17: BREAKING CHANGE to document

orderBy now receives an array of filters

tests/schema/__app/generated/schema.graphql
```diff
@@ -33,8 +33,8 @@ input BubbleMembersWhereInput {
 input BubbleWhereInput {
   AND: [BubbleWhereInput!]
   createdAt: DateTimeFilter
-  id: UUIDFilter
-  members: UserFilter
+  id: StringFilter
```
jasonkuhrt, line 44: BREAKING CHANGE to document

We now [seem to] apply `StringFilter` to model IDs that are of `String` type. This seems sound.

I believe there is an issue about having us generate _not_ `String` type for model IDs but rather GraphQL `ID` types.

When we achieve that, then I would would expect e.g. this to return to something else than `StringFilter`.

For now this actually makes sense and looks more consistent, though.

@Weakky thoughts?

tests/schema/__app/generated/schema.graphql
```diff
@@ -33,8 +33,8 @@ input BubbleMembersWhereInput {
...
+  members: UserListRelationFilter
```
jasonkuhrt, line 45: BREAKING CHANGE to document

The naming pattern here has changed. On the surface, here, it appears to go from `<ModelName>Filter` to `<ModelName>ListRelationFilter`. But maybe the `List` and/or `Relation` parts are variable too.

tests/schema/__app/generated/schema.graphql
```diff
@@ -52,7 +52,7 @@ input DateTimeFilter {
   in: [DateTime!]
   lt: DateTime
   lte: DateTime
-  not: DateTime
+  not: NestedDateTimeFilter
```
jasonkuhrt, line 64: Talked with @timsuchanek. This change reflects dmmf changes made to support https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/filtering#case-insensitive-filtering.

Larger example of what is going on:

```gql
input IntFilter {
  equals: Int
  in: [Int!]
  notIn: [Int!]
  lt: Int
  lte: Int
  gt: Int
  gte: Int
  not: NestedIntFilter
}
input NestedIntFilter {
  equals: Int
  in: [Int!]
  notIn: [Int!]
  lt: Int
  lte: Int
  gt: Int
  gte: Int
  not: NestedIntFilter
}
```

Our conclusion is that at the gql layer we don't need the indirection. We can just have `IntFilter` recurse into itself.

New issue to support case insensitivity in gql layer here https://github.com/graphql-nexus/nexus-plugin-prisma/issues/806.

tests/schema/__app/generated/schema.graphql
```diff
@@ -83,7 +83,54 @@ type Mutation {
...
+  gte: String
+  in: [String!]
+  lt: String
+  lte: String
+  not: NestedStringFilter
+  notIn: [String!]
+  startsWith: String
+}
+
+input NestedStringNullableFilter {
```
jasonkuhrt, line 177: There are also nested filter types for nullable filters.

tests/schema/__app/generated/schema.graphql
```diff
@@ -195,31 +256,15 @@ input UserFilter {
...
+  posts: PostListRelationFilter
 }
 
 input UserWhereUniqueInput {
   id: String
 }
-
-scalar UUID
-
-input UUIDFilter {
```
jasonkuhrt, line 213: BREAKING CHANGE to document

UUID scalar and filter appear to be gone.

jasonkuhrt (commented):
All breaking changes need to be documented in two place:

1. release notes
2. the library docs, especially the "schema contributions" where most of the breaking changes appear to be concentrated.

@janpio for (2) can you please advise where the pr should go to.

tests/schema/__app/generated/schema.graphql
jasonkuhrt: Few points from Dom

- top-level type in dmmf is where `mode` exists
- `mode` is not supported in the nested types that recurse
- this means the types meaningfully diverge
- we cannot put `mode` on every level and have that be usable at the prisma client level
- so to support `mode` we will have to double up our filter types
- so we might as well start now in prep for that later
- there are no features with regards to non-string-ish filters currently
- but there might be in the future, and by way of consistency of pattern it is arguably simpler to have all filter types have nested types.

conclusion is that we will not collapse anything

Instead we need to document this new pattern in the schema contributions docs.

tests/schema/__app/generated/schema.graphql
jasonkuhrt: I added some new models and patterns to the integration test to verify things. It seems the change is as simple as RM_Filter to RM_ListRelationFilter.

tests/schema/__app/generated/schema.graphql
jasonkuhrt: I'm confused why we didn't have this change before since list-ordering by multiple fields comes from https://github.com/prisma/prisma/releases/tag/2.4.1 which we already support.

tests/schema/__app/schema.prisma
```diff
@@ -7,16 +7,24 @@ model Bubble {
...
+  location  Location
   // likes     Post[]
 }
 
+model Location {
+  id        Int @id @default(autoincrement())
+  country   String
+  city      String
+}
+
```
jasonkuhrt, line 27: I made some changes to the integration test to exercise areas of schema generation that I wasn't sure about before.

1. boolean fields and how those show up for e.g. filtering
2. singular (non-list) relations

tests/schema/__app/generated/schema.graphql
jasonkuhrt: I don't think there is really any change here. This diff is just a function of UUID scalar being removed in the first place.

https://github.com/graphql-nexus/nexus-plugin-prisma/pull/805/files#diff-753970daa73034828fbe2f685d2c3f70L4378

jasonkuhrt:
I can't find any place where `isList` was removed in dmmf anymore. Maybe it was mistakenly seen that way before.

jasonkuhrt:
doc pr up https://github.com/graphql-nexus/nexus/pull/1385

Weakky (approved):
🚢

PR #877, opened 2020-10-01
fix: handle json types when converting nulls to undefined

### LINKED ISSUE #797: Could not find input type name: Json
I'm trying to run a migration:

```graphql
mutation UPDATE_DATA($data: Json) {
  updateOneSetting(data:{data: $data} where: {id: 22}) {
    data
  }
}
```

Query variables:

```json
{
  "data": {"custom_date_format": {"type": "advanced", "value": "mm/dd/yyyy"}}
}
```

But I'm getting an error

```bash
INTERNAL_SERVER_ERROR
Error: Could not find input type name: Json at DmmfDocument.getInputTypeWithIndexedFields
```

Here's my schema

```
generator client {
  provider        = "prisma-client-js"
  previewFeatures = ["transactionApi"]
}

datasource db {
  provider = "postgresql"
  url      = env("DATABASE_URL")
}

model Faq {
  id        Int      @default(autoincrement()) @id
  title     String
  body      String
  createdAt DateTime @default(now())
  updatedAt DateTime
}

model Setting {
  id           Int           @default(autoincrement()) @id
  data         Json
  trackingPage TrackingPage?
}

model TrackingPage {
  storeHost        String   @id
  storeId          String?  @unique
  accessToken      String
  storeName        String?
  storeDescription String?
  merchantEmail    String?  @unique
  settingId        Int?     @unique
  settings         Setting? @relation(fields: [settingId], references: [id])
}
```

And objectType

```js
schema.objectType({
  name: 'Setting',
  definition(t) {
    t.model.id()
    t.model.data()
    t.model.trackingPage()
  },
})
```

I took a peek in the generated `nexus.graphql` file and the Json scalar is defined.

Might be related to: https://github.com/graphql-nexus/nexus/issues/1307 and https://github.com/graphql-nexus/nexus-plugin-prisma/issues/723

lvauvillier:
`Json` type is not properly handled when converting nulls to undefined.

fixes #797

### DIFF
--- a/src/null.ts
+++ b/src/null.ts
@@ -30,7 +30,7 @@ export function transformNullsToUndefined(
 
     if (shouldConvertNullToUndefined) {
       graphqlArgs[key] = undefined
-    } else if (Lo.isPlainObject(val)) {
+    } else if (Lo.isPlainObject(val) && prismaArg.inputType.type != 'Json') {
       const nestedPrismaArgs = dmmf.getInputTypeWithIndexedFields(prismaArg.inputType.type).fields
 
       graphqlArgs[key] = transformNullsToUndefined(graphqlArgs[key], nestedPrismaArgs, dmmf)
@@ -39,4 +39,3 @@ export function transformNullsToUndefined(
 
   return graphqlArgs
 }
-


lvauvillier:
The check can be also done on `prismaArg.inputType.kind === 'scalar'` to be more generic

Weakky:
Hey @lvauvillier, I tried reproducing the error laid out in #797 but I couldn't. I believe this was an error due to Prisma not handling json filters for a while, but as you've created this issue 6 days ago I wanted to hear more from you.

What version of Prisma did you use to encounter this issue?

lvauvillier:
@Weakky I use prisma 2.7.0.

The problem is related to the way the nexus plugin recursively converts null to undefined.
There is a recursion testing if the current field is an object or not. In case of Json inputs, even if the field is a scalar, the condition is true and the recursion is done raising the error at the next step (the dmmf mapping doesn't contains the scalar Json key).

Weakky:
Hey @lvauvillier,

Apologies for the late answer. I just reviewed it again and was able to repro it. Didn't realize we were allowed to pass JS objects to the inputs, which was indeed causing issues. I'll merge it now and then add a test in a second PR as it seems I can't push to your branch. Thanks for fixing it 🙏

Issue #875, opened 2020-10-01
missing typegen file in module


JustinVoitel:
### Description
When generateing the typegen from the plugin, the first line in `node_modules/@types/typegen-nexus-plugin-prisma/index.d.ts` 
says: `import * as Typegen from 'nexus-plugin-prisma/typegen'`, but its missing from the module. This results into nexus-plugin-prisma not finding the corresponding types

### Possible fixes
1. either change the path in the typegen in index.d.ts to `nexus-plugin-prisma/dist/typegen/static`
2. or add the file typegen.js to the modules root with content: `module.exports = require('./dist/typegen/static')`

doing either of these above fixes the typings in my project

### Additional Infos
    operating system: windows 10 wsl2
    versions for
        node v14.8.0
        npm 6.14.7
        @nexus/schema 0.16.0
        nexus-plugin-prisma 0.20.0
        @prisma/client & cli 2.7.0

heibel:
Affected as well. Working around it with `sed -i '' 's/nexus-plugin-prisma\/typegen/nexus-plugin-prisma\/dist\/typegen\/static/g' /path/to/output.d.ts`

lewebsimple:
I can confirm changing the generated import to `nexus-plugin-prisma/dist/typegen/static` works with the following versions (as well as GraphQL 14 / 15) :
```
"@nexus/schema": "^0.16.0",
"nexus-plugin-prisma": "^0.20.0",
"@prisma/client": "~2.7.1"
"@prisma/cli": "~2.7.1"
```
The PR I submitted should work without breaking anything I ~~think~~ hope.

Slmii:
idk why but my `@types/typegen-nexus-plugin-prisma/index.d.ts` is being generated in the root of the project and not in `node_modules`. Even if I change the import there it'll auto change back to the old value.

I deleted the `node_modules` and `package-lock.json`, did a fresh install and then run the command `prisma generate`. Im using:
```
"@nexus/schema": "^0.16.0",
"@prisma/client": "2.7.0",
"@prisma/cli": "2.7.0",
"nexus-plugin-prisma": "^0.20.0",
```

PR #879, opened 2020-10-04
fix: missing typegen file in module #875

lewebsimple:
I'm trying to address #875 but I'm having trouble building nexus-plugin-prisma on its own (i.e. outside my project where it's linked).

Is the problem with the generated typegen path more complex than adjusting nexus-plugin-prisma/src/typegen/render.ts ?

### DIFF
--- a/src/typegen/render.ts
+++ b/src/typegen/render.ts
@@ -52,7 +52,7 @@ export function render(params: {
   paginationStrategy: PaginationStrategy
 }) {
   return `\
-import * as Typegen from '${params.nexusPrismaImportId ?? 'nexus-plugin-prisma/typegen'}'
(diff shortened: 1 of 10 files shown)


lewebsimple:
I got my fork of nexus-plugin-prisma to build when it was outside my other project's folder, but using it with `yarn link` gave errors about Prisma client not being generated. I tried replacing the package in node_modules with the built fork, but still no luck. I finally deleted node_modules/nexus-plugin-prisma/node_modules and it worked just fine, the generated file had the proper path for typegen !
So I guess the PR does resolve #875, though I'm not sure if the path should be using `nexus-plugin-prisma/dist/typegen/static` or `nexus-plugin-prisma/src/typegen/static` ...

jasonkuhrt:
Thanks @lewebsimple, closed by #883 just now.

PR #884, opened 2020-10-07
fix: default typegen path

### LINKED ISSUE #873: Typegen artifact generated outside node_modules
Commit f8a017c3bf72243bb699950b779572834e60af2c for version 0.20 renamed many files.
Namely, it moved `src/schema/builder.ts` to `src/builder.ts`.
However, when `shouldGenerateArtifacts` is `true`, the default output path is still `path.join(__dirname, '../../../@types/typegen-nexus-plugin-prisma/index.d.ts')`.
As a result, when I try to generate artifacts, the @types directory is generated in my project's root dir, instead of `node_modules`.
As a temporary workaround, the output path can be overridden through the `outputs.typegen` option.

### LINKED ISSUE #874: Wrong place for generated types definition
Hello there :)

When I use `makeSchema`, the **nexusSchemaPrisma@0.20.0** plugin generate by default typescript definition (@types/typegen-nexus-plugin-prisma/index.d.ts) inside my working directory instead of node_modules. The default path is supposed to be **node_modules/@types/typegen-nexus-plugin-prisma/index.d.ts**

Here is my schema.ts file (it is pretty standard)
```ts
import { makeSchema } from '@nexus/schema'
import { nexusSchemaPrisma } from 'nexus-plugin-prisma/schema'
import * as types from './types'

export const schema = makeSchema({
  types,
  plugins: [
    nexusSchemaPrisma({
      experimentalCRUD: true,
      paginationStrategy: 'prisma',
    }),
  ],
  outputs: {
    schema: true,
  },
  typegenAutoConfig: {
    sources: [
      {
        source: '@prisma/client',
        alias: 'client',
      },
      {
        source: require.resolve('./context'),
        alias: 'Context',
      },
    ],
    contextType: 'Context.Context',
  },
})
```

Weakky:
Closes #873

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -190,7 +190,7 @@ if (process.env.NEXUS_PRISMA_TYPEGEN_PATH) {
 } else if (process.env.LINK) {
   defaultTypegenPath = path.join(process.cwd(), 'node_modules/@types/typegen-nexus-plugin-prisma/index.d.ts')
 } else {
-  defaultTypegenPath = path.join(__dirname, '../../../@types/typegen-nexus-plugin-prisma/index.d.ts')
+  defaultTypegenPath = path.join(__dirname, '../../@types/typegen-nexus-plugin-prisma/index.d.ts')
 }
 
 let defaultClientPath: string


Weakky:
Closes #874

PR #894, opened 2020-10-15
feat(deps): add official support for Prisma 2.9

### LINKED ISSUE #895: Add official support for Prisma 2.9

jasonkuhrt:
closes #895
Since no code changes were required we are keeping official support for 2.8 as well.


package.json
```diff
@@ -45,11 +45,12 @@
...
-    "@prisma/client": "2.8.1",
-    "@prisma/fetch-engine": "2.8.1",
-    "@prisma/get-platform": "2.8.1",
-    "@prisma/migrate": "2.8.1",
+    "@prisma/cli": "2.9.0",
+    "@prisma/client": "2.9.0",
+    "@prisma/fetch-engine": "2.9.0",
+    "@prisma/get-platform": "2.9.0",
+    "@prisma/migrate": "2.9.0",
+    "@prisma/sdk": "2.9",
```
jasonkuhrt, line 53: Not sure why this wasn't there before... not having it broke tests locally.

package.json
```diff
@@ -45,11 +45,12 @@
...
-    "@prisma/client": "2.8.1",
-    "@prisma/fetch-engine": "2.8.1",
-    "@prisma/get-platform": "2.8.1",
-    "@prisma/migrate": "2.8.1",
+    "@prisma/cli": "2.9.0",
+    "@prisma/client": "2.9.0",
+    "@prisma/fetch-engine": "2.9.0",
+    "@prisma/get-platform": "2.9.0",
+    "@prisma/migrate": "2.9.0",
+    "@prisma/sdk": "2.9.0",
```
jasonkuhrt, line 53: Not sure why this wasn't there before... not having it broke tests locally.

tests/__app/generated/nexus-plugin-prisma-typegen.d.ts
```diff
@@ -0,0 +1,165 @@
+import * as Typegen from '../../../src/typegen/static'
```
jasonkuhrt, line 1: We need these generated files for now because tests in CI fail (ts, type errors, static). We can re-think how we manage our tests suites later if we want. Mainly to avoid having to commit these derived files. But until then these are needed.

tests/app.test.ts
```diff
@@ -10,14 +10,7 @@ import { createNexusPrismaInternal, mockConsoleLog } from './__utils'
...
 it('integrates together', async () => {
-  // Setup file system vars & helpers
-  //
-  const projectRoot = path.join(__dirname, '__app')
-
-  const projectReadFile = (relPath: string): Promise<string | undefined> =>
-    fs.readAsync(path.join(projectRoot, relPath))
-
-  const projectPath = (...paths: string[]): string => path.join(projectRoot, ...paths)
+  const fs = fsjp.cwd(path.join(__dirname, '__app'))
```
jasonkuhrt, line 13: Better use of fpjs here.

tests/runtime/ordering.test.ts
```diff
@@ -33,7 +33,7 @@ it('with whitelist', async () => {
 
   const Query = queryType({
     definition(t) {
-      // @ts-ignore
+      // @ts-expect-error
```
jasonkuhrt, line 36: If for some reason `c: true` doesn't raise an error then TS will raise an error because it didn't because of our annotation here. This is an improvement from before where `@ts-ignore` won't complain if its not actually doing anything.

PR #899, opened 2020-10-16
feat: add disableAtomicOperations option

### LINKED ISSUE #830: Option to Disable and Collapse Atomic Operations
[This](https://github.com/prisma/prisma/issues/3491) issue raised with the Prisma team was initiated by ~2.6's introduction of Atomic Operations. As you can see there, I ran into the issue that  "set" was now required with every update/upsert performed on strings/ints/floats - **a breaking change**. I initially believed this was a bug resolved by the 2.6.2 hotfix, however @jasonkuhrt clarified what was in fact happening:

> Hey @pantharshit00, @hatchli; This does not represent a bug in the Nexus Prisma plugin but a limitation with the GraphQL specification which does not support unions in input types (it is coming eventually).
> 
> Nexus Prisma _could_ have a feature that allows you to disable this Prisma feature and "collapse" the atomic operations back to `string` at the GraphQL schema level where a string would infer a `set` (could be configurable too).
> 
> Nexus Prisma _could_ also eventually have rich mapping features that permit ad-hoc customization of how the GraphQL schema maps to the Prisma Client API.
> 
> But as it stands today, if you want to upgrade, you're going to _have_ to do exactly this:
> 
> > go back adding "set" to every single update/upsert field if it's unnecessary.
> 
> You are more than welcome to [make a feature request](https://github.com/graphql-nexus/nexus-plugin-prisma/issues/new) about how you wish the crud system worked instead.

I am hoping that Atomic Operations would be configurable, much in the way CRUD or pagination are currently configurable in nexus-plugin-prisma.

lvauvillier:
GraphQL doesn't support union types. The plugin has to apply a  flattening heuristic and pick by default the broadest member of the union. On update and upsert, the broadest member is the atomic operation, which is not ideal in many cases...
This option disables atomic operations and come back to scalar values.

The default value is false to avoid any breaking change with the latest version

fixes #830

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -158,11 +158,16 @@ export interface Options {
   scalars?: Partial<Record<Typegen.GetGen<'scalars', string>, GraphQLScalarType>>
   computedInputs?: GlobalComputedInputs
   /**
-   * Graphql does not support union types, atomic operation are enabled by default for CRUD update and upsert
+   * Enable atomic operations
    *
-   * @default false
+   * @remarks
+   *
+   * GraphQL doesn't support union types. The plugin has to apply a flattening heuristic and pick by default the broadest member of the union.
+   * On update and upsert, the broadest member is the atomic operation, which is not ideal in many cases.
+   *
+   * @default true
    */
-  disableAtomicOperations?: boolean
+  atomicOperations?: boolean
 }
 
 export interface InternalOptions extends Options {
@@ -262,7 +267,7 @@ export class SchemaBuilder {
       getTransformedDmmf(config.inputs.prismaClient, {
         globallyComputedInputs: this.globallyComputedInputs,
         paginationStrategy: this.paginationStrategy,
-        disableAtomicOperations: config.disableAtomicOperations,
+        atomicOperations: config.atomicOperations,
       })
     this.scalars = (options.scalars as any) ?? {}
     this.publisher = new Publisher(this.dmmf, config.nexusBuilder, this.scalars)
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -226,6 +226,7 @@ const defaultOptions = {
   shouldGenerateArtifacts,
   prismaClient: (ctx: any) => ctx.prisma,
   paginationStrategy: 'relay' as const,
+  atomicOperations: true,
   inputs: {
     prismaClient: defaultClientPath,
   },
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -162,7 +162,7 @@ export interface Options {
    *
    * @remarks
    *
-   * GraphQL doesn't support union types. The plugin has to apply a flattening heuristic and pick by default the broadest member of the union.
+   * GraphQL [doesn't support union input types](https://github.com/graphql/graphql-spec/pull/733). But Prisma Client atomic operations (e.g. [atomic operations on update for Int and Float fields](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/crud#atomic-operations-on-update)) are modelled as a union input type in TypeScript. By default Nexus Prisma will model this as a nested input object type with runtime validation that only one operation is sent by the client. However, you can configure this. If your API clients do not need the power of all the atomic operations then disable them. When disabled the nested input object is replaced by only one of the possible atomic operations. Disabling removes types from your API schema and thus complexity for your clients.
    * On update and upsert, the broadest member is the atomic operation, which is not ideal in many cases.
    *
    * @default true
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -163,7 +163,6 @@ export interface Options {
    * @remarks
    *
    * GraphQL [doesn't support union input types](https://github.com/graphql/graphql-spec/pull/733). But Prisma Client atomic operations (e.g. [atomic operations on update for Int and Float fields](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/crud#atomic-operations-on-update)) are modelled as a union input type in TypeScript. By default Nexus Prisma will model this as a nested input object type with runtime validation that only one operation is sent by the client. However, you can configure this. If your API clients do not need the power of all the atomic operations then disable them. When disabled the nested input object is replaced by only one of the possible atomic operations. Disabling removes types from your API schema and thus complexity for your clients.
-   * On update and upsert, the broadest member is the atomic operation, which is not ideal in many cases.
    *
    * @default true
    */
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -157,6 +157,12 @@ export interface Options {
    */
   scalars?: Partial<Record<Typegen.GetGen<'scalars', string>, GraphQLScalarType>>
   computedInputs?: GlobalComputedInputs
+  /**
+   * Graphql does not support union types, atomic operation are enabled by default for CRUD update and upsert
+   *
+   * @default false
+   */
+  disableAtomicOperations?: boolean
 }
 
 export interface InternalOptions extends Options {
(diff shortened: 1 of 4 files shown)


src/builder.ts
```diff
@@ -157,6 +157,12 @@ export interface Options {
    */
   scalars?: Partial<Record<Typegen.GetGen<'scalars', string>, GraphQLScalarType>>
   computedInputs?: GlobalComputedInputs
+  /**
+   * Graphql does not support union types, atomic operation are enabled by default for CRUD update and upsert
+   *
+   * @default false
+   */
+  disableAtomicOperations?: boolean
```
jasonkuhrt: Remove a level of negation by calling this `atomOperations` and make its default `true`.

src/builder.ts
```diff
@@ -157,6 +157,12 @@ export interface Options {
    */
   scalars?: Partial<Record<Typegen.GetGen<'scalars', string>, GraphQLScalarType>>
   computedInputs?: GlobalComputedInputs
+  /**
+   * Graphql does not support union types, atomic operation are enabled by default for CRUD update and upsert
```
jasonkuhrt: Your PR comment is great, a version of it under `@remarks` would be great too.

```
@remarks

GraphQL doesn't support union types. The plugin has to apply a flattening heuristic and pick by default the broadest member of the union. On update and upsert, the broadest member is the atomic operation, which is not ideal in many cases.
```

jasonkuhrt (changes requested):
Let me know if you're up for a few more things:

1. Tests
2. Docs -> https://github.com/graphql-nexus/schema/blob/develop/docs/content/030-pluginss/050-prisma/030-api.mdx

lvauvillier:
@jasonkuhrt just add tests and a PR in the nexus/schema repo to add some docs.

PS: I don't know why github tests fails and why generated snapshot differs from my local machine...

src/builder.ts
```diff
@@ -157,6 +157,17 @@ export interface Options {
    */
   scalars?: Partial<Record<Typegen.GetGen<'scalars', string>, GraphQLScalarType>>
   computedInputs?: GlobalComputedInputs
+  /**
+   * Enable atomic operations
+   *
+   * @remarks
+   *
+   * GraphQL doesn't support union types. The plugin has to apply a flattening heuristic and pick by default the broadest member of the union.
```
jasonkuhrt:
```suggestion
   * GraphQL [doesn't support union input types](https://github.com/graphql/graphql-spec/pull/733). But Prisma Client atomic operations (e.g. [atomic operations on update for Int and Float fields](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/crud#atomic-operations-on-update)) are modelled as a union input type in TypeScript. By default Nexus Prisma will model this as a nested input object type with runtime validation that only one operation is sent by the client. However, you can configure this. If your API clients do not need the power of all the atomic operations then disable them. When disabled the nested input object is replaced by only one of the possible atomic operations. Disabling removes types from your API schema and thus complexity for your clients.
```

src/builder.ts
```diff
@@ -157,6 +157,17 @@ export interface Options {
...
+   * On update and upsert, the broadest member is the atomic operation, which is not ideal in many cases.
```
jasonkuhrt: Speaking as a user. I don't really understand this. What does "broadest member" mean? Is this point about when atomic operations are `false` or when they are `true`?

jasonkuhrt (changes requested):
Thanks for the tests @lvauvillier! Let's also make sure the jsDoc is great. I started an edit on the first line. Left a comment on the second.

tests/schema/__snapshots__/atomicOperations.test.ts.snap
```diff
@@ -0,0 +1,373 @@
...
+input ModelNameWhereInput {
+  AND: [ModelNameWhereInput!]
+  OR: [ModelNameWhereInput!]
+  NOT: [ModelNameWhereInput!]
+  id: IntFilter
+  name: StringFilter
+  amount: IntFilter
+}
+
+input StringFieldUpdateOperationsInput {
```
jasonkuhrt, line 230: Actually a problem here. Disabling atomic operations should remove the input object nesting.

Was this approach intentional on your part?

jasonkuhrt: Hm, considering what you wrote in the docs here https://github.com/graphql-nexus/schema/pull/587/files#diff-aca15a6355a7aef9e2bd3976c18aa15715429c456cea7531795eabc1d0b572b3 I believe these tests results were not your intention.

lvauvillier: There is two tests in this file: default and disabled. You comment the result when atomic operations are enabled (by default). I can remove this default test and keep only the disabled if this is not clear

tests/schema/__snapshots__/atomicOperations.test.ts.snap
```diff
@@ -0,0 +1,373 @@
...
+  set: String
```
jasonkuhrt, line 231: Where are the other operations like `increment` `decrement` and so on?

lvauvillier: I don't know why these operations are not generated on my local machine. They are generated when tests are running on github. I'll spend more time to understand what happens.

jasonkuhrt: Maybe you have the wrong version of prisma locally?

lvauvillier: I use yarn --frozen-lockfile. This install prisma 2.9.0.
atomicNumberOperations is still a preview feature in 2.9.0 and have to be explicitely enabled in schema (released in 2.10). This can explain why they doesn't appears in the snapshot.

But when i look at those github logs:
https://github.com/graphql-nexus/nexus-plugin-prisma/pull/899/checks?check_run_id=1326841882
With the same commands, atomic operations seems to be properly generated...
I spend two hours to figure out how it can differs without any result.

Can you try to generate the snapshot on your side?

jasonkuhrt: Indeed on my machine the tests as you have them pass:

image

> atomicNumberOperations is still a preview feature in 2.9.0 and have to be explicitely enabled in schema

Yes that is probably the reason.

@timsuchanek or @Weakky maybe you can give a tip here.

1. Tests use a data model of:

```
        model ModelName {
          id      Int @id @default(autoincrement())
          name    String
          amount  Int
        }
```

2. This is passed to https://github.com/graphql-nexus/nexus-plugin-prisma/blob/master/tests/__utils.ts#L23
3. In turn this calls `PrismaClientGenerator.getDMMF`
4. Which is imported from `const PrismaClientGenerator = require('@prisma/client/generator-build')`

This is a dark area of the API to me, not least because its undocumented and typed as `any`.

jasonkuhrt: But since 2.10 Prisma doesn't need preview flags for atomic operations. So the easy thing to do is move to 2.10 here (rebase from trunk once done, soon) and things should just start working.

lvauvillier: I can confirm, after a rebase on master (last commit uses prisma 2.10) all is properly generated 👍 
I can force-push this PR (i dont think github will proprely handle it) or open a new one. What do you prefer?

lvauvillier: Otherwise i can commit here the updated snapshot and the default "squash and merge" will be fine to make it right.

jasonkuhrt: I think you should be able to merge master back on this branch without issue.

Also your idea to manually tweak the the snapshots so that they pass on CI is ok with me too :)

lvauvillier: perfect, i just push the updated snapshot

jasonkuhrt (approved):
LGTM!

PR #916, opened 2020-10-28
feat: warn if peer deps ranges are not satisfied

### LINKED ISSUE #888: Runtime prisma version check
When used with a version of Prisma that does not match the peer dependency version range raise a runtime exception.

Allow a config setting that opts-out of this check: `checks.prismaVersion`.

Weakky:
Closes #888

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,7 @@
 import { colors } from './colors'
+import * as semver from 'semver'
+
+const pkgJson = require('../package.json')
 
 function ensureDepIsInstalled(depName: string) {
   try {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -17,8 +20,34 @@ function ensureDepIsInstalled(depName: string) {
   }
 }
 
+function ensurePeerDepRangeSatisfied(depName: string) {
+  try {
+    const installedVersion = require(`${depName}/package.json`).version
+
+    if (!installedVersion) {
```
jasonkuhrt, line 28:
```suggestion
	// npm enforces that package manifests have a valid "version" field so this case _should_ never happen under normal circumstances.
    if (!installedVersion) {
```

src/index.ts
```diff
@@ -17,8 +20,34 @@ function ensureDepIsInstalled(depName: string) {
...
+      return
+    }
+
+    const supportedRange = pkgJson.peerDependencies[depName]
```
jasonkuhrt:
```suggestion
    const supportedRange = pkgJson.peerDependencies[depName]
    
    if (!supportedRange) {
      console.warn(
        console.red(`The package "nexus-plugin-prisma" has no such peer dep "${depName}" and so the check for if the consumer has satisfied that peer dep cannot be fulfilled.`
        )
      )
      return
    }
```

Since this entire block is inside a try-catch I opted to log here. Probably should be a throw though.

jasonkuhrt (approved):
Just a few polishes, LGTM!

PR #920, opened 2020-10-28
fix: Cannot find module 'semver' 

mehdyouras:
Now that semver is used at runtime it should be placed in dependencies.


Weakky:
Hey @mehdyouras, there was another related mistake that I needed to fix alongside the semver dependency error, so I batched the two in #921. I fixed all of them in one PR, so I'll close yours now. Thanks for taking the time to create a PR though 🙏

PR #965, opened 2020-11-24
feat(deps)!: add support for Nexus@0.19

jasonkuhrt:
BREAKING CHANGE:

Your version of Nexus must be at least v0.19

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "upgrade-prisma-dev": "yarn add @prisma/cli@dev @prisma/client@dev @prisma/fetch-engine@dev @prisma/get-platform@dev @prisma/migrate@dev @prisma/sdk@dev"
   },
   "peerDependencies": {
-    "@nexus/schema": "^0.13.0 || ^0.14.0 || ^0.15.0 || ^0.16.0 || ^0.17.0 || ^0.18.0",
+    "@nexus/schema": "^0.19.0-next.6",
     "@prisma/client": "2.10.x || 2.11.x",
     "graphql": "^15.3.0"
   },
@@ -45,7 +45,7 @@
     "semver": "^7.3.2"
   },
   "devDependencies": {
-    "@nexus/schema": "0.18.0",
+    "@nexus/schema": "^0.19.0-next.6",
     "@prisma-labs/prettier-config": "0.1.0",
     "@prisma/cli": "2.11.0",
     "@prisma/client": "2.11.0",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
     "upgrade-prisma-dev": "yarn add @prisma/cli@dev @prisma/client@dev @prisma/fetch-engine@dev @prisma/get-platform@dev @prisma/migrate@dev @prisma/sdk@dev"
   },
   "peerDependencies": {
-    "@nexus/schema": "^0.13.0 || ^0.14.0 || ^0.15.0 || ^0.16.0 || ^0.17.0 || ^0.18.0",
+    "@nexus/schema": "^0.19.0-next.6",
```
jasonkuhrt: `0.19.x` before this PR gets merged

src/publisher.ts
```diff
@@ -179,3 +179,32 @@ export class Publisher {
...
+  if (fieldType.isList) {
+    return [Nexus.nonNull, Nexus.list, Nexus.nonNull]
+  } else if (fieldType.isRequired === true) {
+    return [Nexus.nonNull]
+  } else if (fieldType.isRequired === false) {
+    return [Nexus.nullable]
+  }
+
+  return [] as Function[]
+}
```
jasonkuhrt, line 210: In these functions the order you see here is as you would with `t.*` or `a(b(c()))`.

That is why above we `reduceRight`, rather than `reduce`.

src/publisher.ts
```diff
@@ -179,3 +179,32 @@ export class Publisher {
...
 }
+
+/**
+ * Get the composition-order pipeline of Nexus type def functions that will match the nullability and list types of the Prisma field type.
+ *
+ * For example { isList:true, isRequired: true } would result in array of funcs: [nonNull, list, nonNull]
+ */
+export const getNexusTypesCompositionForInput = (fieldType: DmmfTypes.SchemaArg['inputType']) => {
+  if (fieldType.isList) {
+    return [Nexus.list, Nexus.nonNull]
```
jasonkuhrt, line 192: In these functions the order you see here is as you would with `t.*` or `a(b(c()))`.

That is why above we `reduceRight`, rather than `reduce`.

This point is only actually applicable to this case so leaving the comment on this line.

(all other cases are only one type augmentation or symmetrical on both sides)

tests/__app/generated/nexus-typegen.d.ts
```diff
@@ -340,58 +340,16 @@ export interface NexusGenRootTypes {
   }
 }
 
-export interface NexusGenAllTypes extends NexusGenRootTypes {
-  BoolFilter: NexusGenInputs['BoolFilter'];
```
jasonkuhrt, line 344: I don't understand what's going on here... normal?

jasonkuhrt:
For some reason the old version of Nexus is being used in CI... build works fine for me locally.

Confused, will try tomorrow.

image

PR #985, opened 2020-12-09
feat(deps): add support for Prisma 2.13

jasonkuhrt:
closes #956

No breaking changes to the generated GraphQL API detected.

BREAKING CHANGE:

Peer dependency requirement for `@prisma/client` raised from 2.12 to 2.13

No actual breaking changes beyond this are known, but we're still labelling this as a breaking change as it technically introduces backwards incompatible package requirements.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "peerDependencies": {
     "@nexus/schema": "0.19.x",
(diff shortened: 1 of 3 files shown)


Weakky (approved):
LGTM 👍

PR #1038, opened 2021-01-26
fix: common options to CRUD typings

### LINKED ISSUE #1037: Typings to allow to pass common options in CRUD operations
Allow to pass the common options to the CRUD field.

The implementation here already allow us to use the common option `description` & `deprecation`:
https://github.com/graphql-nexus/nexus-plugin-prisma/blob/567194c861d5093c18dc412a5564606b23ba8484/src/builder.ts#L561-L581

The problem lies with the typings which only allows to pass the options `pagination`, `ordering`, `filtering`, `type`, `alias` , `resolve` & `computedInput`.

haleksandre:
Add CommonFieldConfig to BaseRelationOptions to allow to pass common options such as `description` & `deprecation` to the CRUD configuration.

This should fix #1037

### DIFF
--- a/src/typegen/static.ts
+++ b/src/typegen/static.ts
@@ -169,7 +169,8 @@ type BaseRelationOptions<
    */
   resolve?: CustomFieldResolver<TypeName, Alias extends undefined ? MethodName : Alias>
   computedInputs?: LocalComputedInputs<MethodName>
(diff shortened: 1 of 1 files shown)


jasonkuhrt (approved):
ty!

PR #1048, opened 2021-02-11
feat: export BaseRelationOptions type

hajnalben:
When writing extensions to work with nexus-plugin-prisma we need this type to be exported.


jasonkuhrt (approved):
sure!

PR #1125, opened 2021-11-22
fix: Make addComputedInputs work when data is undefined

### LINKED ISSUE #678: "Cannot convert undefined or null to object" when running an upsert or delete mutation
My `t.crud.upsertOneXXXX` and `t.crud.deleteOneXXXX` operations are throwing this error:
```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at addGloballyComputedInputs (...\node_modules\nexus-prisma\src\dmmf\transformer.ts:146:17)
    at Object.addComputedInputs (...\node_modules\nexus-prisma\src\dmmf\transformer.ts:177:10)
    at resolve (...\node_modules\nexus-prisma\src\builder.ts:301:28)
    at field.resolve (...\node_modules\graphql-extensions\src\index.ts:155:61)
    at resolveFieldValueOrError (...\node_modules\graphql\execution\execute.js:467:18)
    at resolveField (...\node_modules\graphql\execution\execute.js:434:16)
    at ...\node_modules\graphql\execution\execute.js:244:18
    at ...\node_modules\graphql\jsutils\promiseReduce.js:23:10
    at Array.reduce (<anonymous>)
```

It seems that at `\node_modules\nexus-prisma\src\dmmf\transformer.ts` line 181 it does `params.args.data`. `data` doesn't exist when we are deleting an entity nor when we're upserting. Upserting uses `create` and `update` args but not a `data` one, and delete has none.

This also happens with deleteMany. Update and create works fine.

### LINKED ISSUE #988: computedInputs option causes an error on mutations without `data` arg
Hi, this line on `addGloballyComputedInputs` throws `TypeError: Cannot convert undefined or null to object` on any mutation that doesn't have a `data` arg, like `crud.upsert` and `crud.delete`

https://github.com/graphql-nexus/nexus-plugin-prisma/blob/6c8801c6e1d99bfdb73a7c1c89db9607712b0e01/src/dmmf/transformer.ts#L227

rhysawilliams2010:
Fix issue where computed inputs throw an exception for mutations which don't pass data. (i.e. delete mutations)

fixes #678 
fixes #988

### DIFF
--- a/src/dmmf/transformer.ts
+++ b/src/dmmf/transformer.ts
@@ -222,6 +222,7 @@ async function addGloballyComputedInputs({
     }),
     Promise.resolve({} as Record<string, any>)
   )
+
   // Combine computedInputValues with values provided by the user, recursing to add
   // global computedInputs to nested types
   return Object.keys(data).reduce(async (deeplyComputedData, fieldName) => {
@@ -248,6 +249,10 @@ export async function addComputedInputs({
   locallyComputedInputs,
   params,
 }: AddComputedInputParams) {
+  if (params.args.data === undefined) {
+    return { ...params.args }
+  }
+
   return {
     ...params.args,
     data: {
--- a/tests/schema/__snapshots__/computedInputs.test.ts.snap
+++ b/tests/schema/__snapshots__/computedInputs.test.ts.snap
@@ -7,6 +7,7 @@ Object {
 }
 
 type Mutation {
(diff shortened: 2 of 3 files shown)


willdunlop (approved):

rhysawilliams2010:
This only stops the delete mutation working. Upsert is still broken :(

